PR #178, opened 2014-04-06
Add incremental forward and reverse search

mkleehammer:
This adds incremental forward (cmd-i) and reverse (cmd-shift-i) searches.

### DIFF
--- a/keymaps/find-and-replace.cson
+++ b/keymaps/find-and-replace.cson
@@ -14,12 +14,15 @@
   'cmd-ctrl-g': 'find-and-replace:select-all'
   'cmd-d': 'find-and-replace:select-next'
   'cmd-e': 'find-and-replace:use-selection-as-find-pattern'
+  'cmd-i': 'incremental-find:forward'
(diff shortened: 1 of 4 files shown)


benogle:
So there is now a setting to select the first match:

screen shot 2015-08-11 at 4 48 00 pm

How about making `cmd-i` and `cmd-shift-i` basically do the same thing as `cmd-g` but keep focus on the 
find and replace pattern. Would that solve your issue?

MartinMuzatko:
isn't cmd-shift-i opening the dev tools?

maxbrunsfeld:
@mkleehammer Thanks for your work on this; I'm going to close this one out until you get a chance to come back to it. FYI, we're planning on making big changes to this package soon to remove jQuery, so further conflicts will be created.

mkleehammer:
Sorry, I must have missed responding.  I'm afraid I'm using Spacemacs lately so my opinion shouldn't really count anymore.  Good luck!

Issue #67, opened 2013-10-31
Sublime functionality Atom doesn't support


probablycorey:
* [x] `'meta-d': 'find_under_word'`
* [x] `'meta-k meta-d': 'find_under_expand_skip'`
* [x] `'meta-alt-e': 'replace_next'`
* [x] `'meta-shift-e': 'slurp_replace_string'`
* [x] `'alt-meta-g': 'find_under'`
* [x] `'shift-alt-meta-g': 'find_under_prev'`
* [ ] `'ctrl-meta-g': 'find_all_under'`
* [x] `'enter': 'find_next', 'context': [{'key': 'panel', 'operand': 'find'}, y': 'panel_has_focus'}] },`
* [x] `'shift-enter': 'find_prev', 'context': [{'key': 'panel', 'operand': d'}, {'key': 'panel_has_focus'}] },`
* [x] `'alt-enter': 'find_all', 'args': {'close_panel': true}, 'context': ey': 'panel', 'operand': 'find'}, {'key': 'panel_has_focus'}] },`
* [x] `'ctrl-alt-enter': 'replace_all', 'args': {'close_panel': true}, text': [{'key': 'panel', 'operand': 'replace'}, {'key': 'panel_has_focus'}] },`

grigio:
+1 to find_all

grigio:
@clhynfield thanks, I've upgraded to 0.166 and it work well

tjmehta:
`meta-d` does not work as it does in Sublime
In sublime, if I select anything include partial word, or any segment of text,
`meta-d` will select the next match.
In atom, `meta-d` will only match exact words.
Ex:
```js
var personName = 'name';
var personRole = 'admin';
```
I should be able to highlight `person` (of `personName`), and  `meta-d` should select 
the next `person` (of `personRole`). `meta-g` on the other hand works as it (and `meta-d`) 
finding the next match.

fnkr:
+1 for find_all

s2t2:
aah i see "find all" is equivalent to: `ctrl-f`, then type the search term, then hit `alt-enter`

thanks!

https://github.com/atom/find-and-replace/issues/67#issuecomment-69021689

bronson:
@probablycorey, looks like you can cross 'shift-enter' off.  It works great for me on 1.0.19.

ruthgrace:
What is the difference between find_under and find_next ? I'm interested in contributing to Atom with this beginner issue.

[...]

PR #229, opened 2014-06-07
Add shift-enter to show previous match

brendankemp:
- Keybinding only active in the find editor field
- Matches Sublime finder functionality

Ref: https://github.com/atom/find-and-replace/issues/67


head898:
#1

head898:
#229

PR #251, opened 2014-07-19
Added folding for project find results

### LINKED ISSUE #98: Let me fold results
We used to have the "expand / collapse" button. I'm guessing this was taken out for a reason, but I found it very useful in, for example, hiding results from a _test_ folder.

DepthDeluxe:
Allows users to fold results in the project fine result view, hiding all results for a single file.  The folding can be done either by clicking on the file in the results list or pressing the left and right arrow keys.

out

resolves #98


lepregh:
Should "File > Folding" menu work on this?

Issue #195, opened 2014-05-05
Match whole word option.


yglukhov:
Please, implement a "match whole (entire) word" option, like it's done in lots of other text editors.

PR #260, opened 2014-07-28
Feature: whole word option

abe33:
As requested in #195

Many things are open to suggestions, like the option button label or the key bindings.
- [x] Implement the whole-word search.
- [x] Handle serialization of the whole-word option state.
- [x] Add an option button and adjust styles of the group so that it can fit in it.
- [x] Display the whole-word option state in the label.

Don't hesitate to comment if you see anything that is missing.

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -370,6 +370,7 @@ class FindView extends View
     else
       label.push('Case Insensitive')
     label.push('Within Current Selection') if @findModel.inCurrentSelection
+    label.push('Whole Word') if @findModel.wholeWord
     @optionsLabel.text(label.join(', '))
 
   toggleRegexOption: =>


jeffschwartz:
Don't see this find in project. Should be imo. Makes refactoring whole project easier.

ArnaudRinquin:
Same, I don't know where to access it. Any idea?

lee-dohm:
1. From the menu: `Find > Find in Project`
1. From the Command Palette: `Project Find: Toggle`
1. On OS X you can use the keybinding <kbd>Shift+Cmd+F</kbd> (I believe it is <kbd>Ctrl+Shift+F</kbd> on other platforms)

jeffschwartz:
I'm sorry, perhaps I should have been more specific. I don't see the option to 'find by whole words' for project find.

lee-dohm:
Ah ok, sorry ... my eyes must still be sleepy :grinning:

It seems this only added it for non-project find. Would you mind opening a new issue?

jeffschwartz:
Sure, no problem.

PR #311, opened 2014-11-21
Upgrade to the new APIs

benogle:
The goal is to get rid of all the deprecations. 
### TODO
- [x] Make buffer find specs pass on new space-pen
- [x] Make project find specs pass on new space-pen
- [x] No deprecations in buffer find
- [x] No deprecations in project find


PixleProgrambetaxl:
+1

PR #314, opened 2014-12-02
Find all

benogle:
This is #290 but rebased against the current master, and with some cleanup.

Closes #290


qoomon:
First thank you so much for this feature.
However it does not work with enabled "Only In Selection" search option.

Issue #332, opened 2014-12-19
use-selection workflow is disruptive, unwieldy


clhynfield:
Using keyboard shortcuts in [other editors](https://github.com/atom/find-and-replace/issues/67) to set Find pattern, select next/all, and replace is smooth and efficient — select text, ⌘E, ⌘⌃G, and type to replace: 

![sublime-use-selection](https://cloud.githubusercontent.com/assets/148868/5504809/fb4557a8-8756-11e4-951c-eb7481fb1957.gif)

In Atom, use-selection-as-find-pattern brings up the find panel, requiring the user to dismiss it or switch focus back to the editor before using select-next or select-all keyboard shortcuts: 

![atom-use-selection](https://cloud.githubusercontent.com/assets/148868/5504835/51b0e42c-8757-11e4-9912-ffdab87208a9.gif)

Showing the panel is right for ⌘F, and grabbing the current selection as the find pattern on show is a handy default action, but for a sleek use-selection and select next/all workflow, there's got to be a smoother way.

izuzak:
@clhynfield I'm able to do this without showing the panel. Here's a gif:

![select](https://cloud.githubusercontent.com/assets/38924/6170006/a4c91382-b2d1-11e4-800f-37152ace4fe2.gif)

Does that help?

clhynfield:
Thanks, @izuzak, that does help a bit—I didn't know you could find all occurrences without setting the find pattern. Still, that doesn't work for find-next or find-previous, does it?

izuzak:
@clhynfield You're right. I was just pointing out that there's a better way to do what you did in the GIFs you recorded.

Also, clarifying what exactly is possible and what exactly isn't possible or how exactly it could be improved is always helpful in pushing issues forward. Some users reading the issue don't have the same context as you do and might not know what you said about OS X or other editors in your recent comment. Sharing as many details as possible in the original issue and doing so in a clear way is _always_ helpful. :blue_heart:

schmurfy:
this is really annoying, I am used to the behavior @clhynfield demonstrated and I find the actual atom behavior really disturbing, the example was maybe not the best but here is what I use cmd+e for:
- quickly replace a word in multiple lines (cmd+e, cmd+g, type new text)
- replace all (cmd+e, ctrl+cmd+g, type text)

With Atom currently this becomes: cmd+e,escape,cmd+g, type text.
What was the idea behind the current behavior ? is there actually someone who finds it ok ?

[...]

PR #333, opened 2014-12-19
Don't show Find panel on use-selection

clhynfield:
use-selection-as-find-pattern still does what it says on the tin, but much
more efficiently for keyboard users. Fixes atom/find-and-replace#332.

All specs still pass.


brkattk:
:+1:

benogle:
Closing as opening the find panel highlights all the matches and give the user feedback that something has happened.

schmurfy:
I just found this, could we at least have an option to toggle this off ?
Haivng the panel pop up when I use cmd+e is really annoying.

PR #344, opened 2015-02-02
setSelectionAsFindPattern updates

adzenith:
These update the <kbd>cmd+e</kbd> behavior to use the word under the cursor if no word is selected (like <kbd>cmd+d</kbd> does) and to not clear the find pattern if nothing's selected and there's no word under the cursor.

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -329,7 +329,7 @@ class FindView extends View
       editor.scrollToCursorPosition(center: true)
 
   setSelectionAsFindPattern: =>
-    pattern = @findModel.getEditor()?.getSelectedText()
+    pattern = @findModel.getEditor()?.getSelectedText() or @findModel.getEditor()?.getWordUnderCursor()
     @updateModel {pattern} if pattern
 
   updateOptionsLabel: ->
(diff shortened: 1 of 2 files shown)


benogle:
Can you add specs for this?

lib/find-view.coffee
```diff
@@ -329,8 +329,9 @@ class FindView extends View
       editor.scrollToCursorPosition(center: true)
 
   setSelectionAsFindPattern: =>
-    pattern = @findModel.getEditor().getSelectedText()
-    @updateModel {pattern}
+    pattern = @findModel.getEditor()?.getSelectedText() or @findModel.getEditor()?.getWordUnderCursor()
+    if pattern
+      @updateModel {pattern}
```
benogle: This should be 

```
@updateModel({pattern}) if pattern
```

lib/project-find-view.coffee
```diff
@@ -266,6 +266,6 @@ class ProjectFindView extends View
 
   setSelectionAsFindPattern: =>
     editor = atom.workspace.getActivePaneItem()
-    if editor?.getSelectedText?
-      pattern = editor.getSelectedText()
+    pattern = editor?.getSelectedText() or editor?.getWordUnderCursor()
+    if pattern
       @findEditor.setText(pattern)
```
benogle: Same. Postfix on the if

adzenith:
Whoops! Forgot about those inline ifs. Thanks.

I might look into adding some specs, but it's crunch time at work so it wouldn't be for a while.

adzenith:
Closing. Superseded by #376.

PR #356, opened 2015-02-27
Removes unnecessary check for non-word character boundaries in selectNextOccurrence

### LINKED ISSUE #340: Select next misses matches when starting with selection at the end of a word
I've noticed that when my selection ends at the end of a word, as I `cmd+d` through matches it skips matches that don't also end a word. I expect that it would select all matching text.

example screencap
In the above example, select next misses the match within the string `contains_tags_scope`. I presume this is due to the underscore (a regex word character), but I expect it to select that part of the text regardless.

brkattk:
Fixes #340 

TL;DR - If I've selected a string of text, I expect that I'm able to jump to every occurrence of that string in the buffer, regardless of it's boundaries. 

Note: There is a failing test. For the life of me, I can't figure out why it is failing - it should work. Maybe it's just a stupid mistake because I'm tired.

### DIFF
--- a/lib/select-next.coffee
+++ b/lib/select-next.coffee
@@ -82,11 +82,6 @@ class SelectNext
     selection = @editor.getLastSelection()
     text = _.escapeRegExp(selection.getText())
 
-    @wordSelected ?= @isWordSelected(selection)
-    if @wordSelected
-      nonWordCharacters = atom.config.get('editor.nonWordCharacters')
-      text = "(^|[ \t#{_.escapeRegExp(nonWordCharacters)}]+)#{text}(?=$|[\\s#{_.escapeRegExp(nonWordCharacters)}]+)"
-
     @editor.scanInBufferRange new RegExp(text, 'g'), range, (result) ->
       if prefix = result.match[1]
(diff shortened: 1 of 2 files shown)


benogle:
What happens when you put a `forgot` in that spec? Does it match the `for` in `forget`?

brkattk:
Not sure which spec you're referring to, but if I have selected `for` it should match any instance of `for` regardless of whether it's part of a word or not.

benogle:
I'm inclined to keep it the way it is. Maybe the regex that handles the boundary should be adjusted for the `_` case. Maybe the `nonWordRegex` setting should not be used here. But when I search for `for`, I do not want it to match `[for]get`.

brkattk:
I'm open to whatever everyone thinks is best. I personally find it frustrating that the matches that [richrace/highlight-selected](https://github.com/richrace/highlight-selected) finds do not match the find-and-replace matches (and also other editors - from Sublime to vim and less [linux]). It's tripped me up a few times, thinking I replaced everything I needed to, only to find there were weird cases like with underscores.

Issue #358, opened 2015-03-05
Quirky behavior with small layouts/window size


LouisTakePILLz:
The `find-and-replace` pane does not seem to adjust too well with small layouts.

Well, enough said; here's a picture showing the erratic behavior:

![atom-fnr small layout-quirky behavior](https://cloud.githubusercontent.com/assets/920910/6496298/55362a92-c2a0-11e4-8ef7-1b2f73a6f0be.gif)

PR #359, opened 2015-03-05
Prevent search field from collapsing when layout space is insufficient

LouisTakePILLz:
The changes that I brought to the LESS stylesheet aim to fix the issue mentioned in #358.

Line-wrapping fix|Overflow fix
-----------------------|----------------
atom-fnr line wrap|atom-fnr tiny layout

Oh, and that also prevents the `.find-meta-container` from overflowing and blocking interaction with the search field.

### DIFF
--- a/styles/find-and-replace.less
+++ b/styles/find-and-replace.less
@@ -141,6 +141,7 @@ atom-workspace.find-visible {
     z-index: 2;
     font-size: .9em;
     line-height: @component-line-height;
+    pointer-events: none;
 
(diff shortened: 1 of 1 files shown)


simurai:
@LouisTakePILLz Thanks for the fix! It will cut off `box-shadow`s, but maybe that shouldn't be the package's concern.

I wonder if there is a way to make the text field wrap to its own row when space gets narrow. I'll give that a try. Really wish element queries would exist.

PR #361, opened 2015-03-06
Wrap UI on narrow width

simurai:
This is a continuation of PR #359.

If the width gets too narrow, the mini editors wrap to a new row and the buttons stretch full width. Not that likely people use it that narrow, but still looks less broken.

Before:

screen shot 2015-03-06 at 9 54 02 am

After:

far

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -8,37 +8,36 @@ module.exports =
 class FindView extends View
   @content: ->
     @div tabIndex: -1, class: 'find-and-replace', =>
-      @div class: 'block', =>
-        @span outlet: 'descriptionLabel', class: 'description', 'Find in Current Buffer'
-        @span class: 'options-label pull-right', =>
+      @div class: 'info-block', =>
+        @span outlet: 'descriptionLabel', class: 'info-block-item description', 'Find in Current Buffer'
+        @span class: 'info-block-item options-label pull-right', =>
           @span 'Finding with Options: '
           @span outlet: 'optionsLabel', class: 'options'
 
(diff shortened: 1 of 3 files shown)


benogle:
Waaay better :+1:

simurai:
Ok, I think it's ready.

Looking at the gif.. the status-bar could also need some overlapping fix.

LouisTakePILLz:
This seems to break the layout on my end, regardless of the theme or the available width:

atom-fnr broken layout

simurai:
@LouisTakePILLz Very odd, can't reproduce it. Looks like everything keeps being wrapped. Are you on Windows? Might be a Chromium bug.

> I have also noticed that the horizontal scrollbar gets triggered at some point, in the parent container.

Yes, that's intended. There is a `min-width`. So when very narrow it starts to scroll.

LouisTakePILLz:
Yup, I'm on Windows 7 (64-bit).

I've also tried finding a way around this, but no luck so far.
Have you considered wrapping both the `.btn-group-find` and `.btn-group-options` in the same div, so it doesn't wrap?

simurai:
Yes, the `.btn-group-find` and `.btn-group-options` have an own div.

screen shot 2015-03-07 at 1 20 30 am

In my case, they always stay next to each other and don't wrap:

find-and-replace

Anyways, might be an issue with the Windows version. I'll try to test it there tomorrow.

LouisTakePILLz:
Everything's sorted out. It turns out I was having issues with my development environment; it's looking much better now!

simurai:
:+1: ok, glad it's not some weird bug.

PR #368, opened 2015-03-21
Add setting to disable livesearch

Afforess:
I added a setting for livesearch. When enabled (by default, it is enabled) the existing behavior for searching occurs.

When disabled, the search waits a small amount of time (750ms seemed reasonable in my tests), and checks to see if the user input any further search text. If so, the model is not updated. If no additional text was entered, the search results are highlighted.

A possible enhancement might be to detect if the user is editing a large file and automatically disabling live search for only that file.

Sorry if the code is not perfect, I have no previous coffeescript experience.

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -177,8 +177,15 @@ class FindView extends View
     @findPrevious(focusEditorAfter: atom.config.get('find-and-replace.focusEditorAfterSearch'))
 
   liveSearch: ->
-    pattern = @findEditor.getText()
-    @updateModel { pattern }
+    if atom.config.get('find-and-replace.liveSearch')
+      pattern = @findEditor.getText()
+      @updateModel { pattern }
+    else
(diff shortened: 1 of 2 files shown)


bronson:
Just curious, why do you want to disable livesearch?  Just performance?

Afforess:
Yes, performance. On large files, live search can bring my machine to a crawl, and I have a fairly powerful desktop. Disabling live search does not disable the in-text search, but does defer it for a short time, which allows for all the search term be entered before the actual search begins. It is still fairly "live" and improves large file search performance significantly.

Afforess:
Yes, performance. On large files, live search can bring my machine to a crawl, and I have a fairly powerful desktop. Disabling live search does not disable the in-text search, but does defer it for a short time, which allows for all the search term be entered before the actual search begins. It is still fairly "live" and improves large file search performance significantly.

PR #375, opened 2015-04-02
Fix issue with interface breaking in small windows

peterjeliot:
This fixes an issue and a related inconsistency in the display of the text at the top of the find panel, that occurs when the window is small enough that not all of the text fits.

There are other solutions (like letting the text overlap), but I think mine is not horribly intrusive, still readable, and fixes the issue at hand of the search box being broken by the floating text wrapping.

Before:
screen shot 2015-04-01 at 7 03 13 pm
screen shot 2015-04-01 at 7 03 01 pm

After:
screen shot 2015-04-01 at 8 49 54 pm
screen shot 2015-04-01 at 8 50 05 pm

### DIFF
--- a/lib/project-find-view.coffee
+++ b/lib/project-find-view.coffee
@@ -13,11 +13,11 @@ module.exports =
 class ProjectFindView extends View
   @content: ->
     @div tabIndex: -1, class: 'project-find padded', =>
-      @div class: 'block', =>
+      @div class: 'find-header block', =>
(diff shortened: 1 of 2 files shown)


benogle:
cc @simurai I thought you had a fix for this a while ago??

simurai:
@benogle Right. Was #361. Wasn't merged yet.

@peterjeliot Thanks a lot for your PR. :bow: #361 should also fix your issue. Should be in the next release.

Issue #210, opened 2014-05-16
Find next/previous occurence of word under cursor


thomasthorsen:
This is a request to add functionality similar to find-next/find-previous but automatically pick up the word under the cursor as the current search term (and don't open the search box). Subsequent uses of find-next/find-previous will also use this word. The natural key-binding is ctrl-f3 and ctrl-shift-f3 (for Linux at least, but also Windows assuming that find-next/find-previous is moved to f3/shift-f3 which is very common in Windows programs (including Sublime Text))

lazutkin:
:+1:

benogle:
Closed by #376

PR #376, opened 2015-04-02
Implement find-next/previous-selected functionality

thomasthorsen:
Add ability to search for the next and previous occurrence of the
current selection or the word containing the cursor in the absence
of a selection. This behaviour, and associated keybindings, is
identical to that of Sublime Text and Visual Studio, among others.

This change also makes the use-selection-as-find-pattern
command use the word containing the cursor in the absence of a
selection instead of simply clearing the existing search pattern, so
that the behaviour is consistent (and code can be reused internally).

This change implements the enhancement described in issue #210.

### DIFF
--- a/keymaps/find-and-replace.cson
+++ b/keymaps/find-and-replace.cson
@@ -11,6 +11,8 @@
 '.platform-darwin atom-text-editor':
   'cmd-g': 'find-and-replace:find-next'
   'cmd-G': 'find-and-replace:find-previous'
+  'cmd-f3': 'find-and-replace:find-next-selected'
+  'cmd-shift-f3': 'find-and-replace:find-previous-selected'
   'cmd-ctrl-g': 'find-and-replace:select-all'
   'cmd-d': 'find-and-replace:select-next'
   'cmd-e': 'find-and-replace:use-selection-as-find-pattern'
@@ -20,6 +22,8 @@
 '.platform-win32 atom-text-editor, .platform-linux atom-text-editor':
   'f3': 'find-and-replace:find-next'
   'shift-f3': 'find-and-replace:find-previous'
+  'ctrl-f3': 'find-and-replace:find-next-selected'
+  'ctrl-shift-f3': 'find-and-replace:find-previous-selected'
   'alt-f3': 'find-and-replace:select-all'
   'ctrl-d': 'find-and-replace:select-next'
   'ctrl-e': 'find-and-replace:use-selection-as-find-pattern'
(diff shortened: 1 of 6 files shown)


lib/find-view.coffee
```diff
@@ -329,9 +331,19 @@ class FindView extends View
       editor.scrollToCursorPosition(center: true)
 
   setSelectionAsFindPattern: =>
+    unless @findModel.getEditor().getSelectedText()
+        @findModel.getEditor().selectWordsContainingCursors()
```
50Wliu: Extra indentation here?

50Wliu:
Could you please add some specs for this?

thomasthorsen:
I will look into adding specs but I'll have to figure out how it works first, so might take me a while. 

It might also be worth noting that these change also implements the changes of pull request #344

lib/find-view.coffee
thomasthorsen: You're right, it should only be 2 spaces for each indent. It is fixed now.

thomasthorsen:
Suggestions for Mac key bindings are welcome!

thomasthorsen:
There is now a full suite of specs for the added functionality including the enhancement of the find-and-replace:use-selection-as-find-pattern command. This is ready to go in if you're happy with the changes. Let me know if there is some further work that needs to be done.

thomasthorsen:
I've rebased to master, can you please consider merging this pull request so I don't have to keep maintaining it. If there are some issues with it please let me know.

keymaps/find-and-replace.cson
```diff
@@ -20,6 +20,8 @@
 '.platform-win32 atom-text-editor, .platform-linux atom-text-editor':
   'f3': 'find-and-replace:find-next'
   'shift-f3': 'find-and-replace:find-previous'
+  'ctrl-f3': 'find-and-replace:find-next-selected'
```
lee-dohm: Probably good to add these to the `.platform-darwin` keymap too as <kbd>cmd-f3</kbd> and <kbd>cmd-shift-f3</kbd>.

thomasthorsen: Done

thomasthorsen:
Any other comments/feedback preventing this from being merged?

thomasthorsen:
Can you please merge this or reject it or give some kind of indication about what the plan is?

benogle:
Thanks for the contribution, blending in, and the ping. Apologies for the delay. :heart:

PR #377, opened 2015-04-08
Add coffeelint support

mnquintana:
Refs atom/atom#6230, atom/atom#5982

### DIFF
--- a/lib/project/results-model.coffee
+++ b/lib/project/results-model.coffee
@@ -67,9 +67,6 @@ class ResultsModel
   onDidRemoveResult: (callback) ->
     @emitter.on 'did-remove-result', callback
 
-  onDidFinishSearching: (callback) ->
-    @emitter.on 'did-finish-searching', callback
-
   serialize: ->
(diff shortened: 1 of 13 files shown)


lib/project/results-model.coffee
```diff
@@ -67,9 +67,6 @@ class ResultsModel
   onDidRemoveResult: (callback) ->
     @emitter.on 'did-remove-result', callback
 
-  onDidFinishSearching: (callback) ->
```
kevinsawicki: Hmm, any reason to remove this? Was it just unused?

mnquintana: It's a duplicate - it's already defined at https://github.com/mnquintana/find-and-replace/blob/coffeelint/lib/project/results-model.coffee#L46-L47

kevinsawicki: Whoa, nice to have this catch that :+1:

kevinsawicki:
Thanks!

PR #385, opened 2015-05-05
Only show scrollbar when there is overflow

mthadley:
There seems to be an always present scrollbar at the bottom of the panel, you can see it here on two different themes (atom-dark-ui and outlander-ui):

before-1
before-2

Its still there on other themes like one-dark, however it is hard to see since the `::--webkit
-scrollbar-track` has the same background-color as it's container. Here is what it looks like with this change: 

after-1
after-2

The scrollbar still appears when the content reaches the `@min-width: 200px`, so that part will still work. 

This is all assuming that we don't want to want to always show the scrollbar track. If `overflow-x` was set to `scroll` intentionally for this reason, than I suppose this change can be disregarded. However, I would say that it looks much better this way. 

Thanks, and let me know if I missed anything!

Note: Reproduced/Tested on OSX (10.10.3) and Linux (Arch + Gnome 3.16).

### DIFF
--- a/styles/find-and-replace.less
+++ b/styles/find-and-replace.less
@@ -45,7 +45,7 @@ atom-workspace.find-visible {
 
   -webkit-user-select: none;
   padding: @component-padding/2;
-  overflow-x: scroll;
+  overflow-x: auto;
 
   .header {
     padding: @component-padding/4 @component-padding/2;


simurai:
:+1:  Great suggestion.

PR #390, opened 2015-05-15
Replace icons 

### LINKED ISSUE #370: Find + Replace icons
Would be nice if the find + replace buttons would have graphical icons instead of text. Related: #231

screen shot 2015-03-26 at 8 54 08 am

Probably should be part of the Octicons, but not sure yet if it's better to fork+extend or create a second Atom specific set.

simurai:
This PR replaces the text icons with SVG versions. Closes #370

Before:

screen shot 2015-05-16 at 12 11 42 am

After:

screen shot 2015-05-16 at 12 08 54 am

### Concerns

Not sure about adding the SVG inside the find-view: https://github.com/atom/find-and-replace/blob/19b742461795b0a19c03ff8ce076f0f2249c6c03/lib/find-view.coffee#L46-L76 It needs to be inline so that the icon color changes based on the text color. But would be better if it can stay an external file. Does space pen have some way to do an "include" instead of `@raw`? Like:

```coffee
@include 'atom://find-and-replace/assets/icons.svg'
```

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -24,10 +24,14 @@ class FindView extends View
           @div class: 'btn-group btn-group-find', =>
             @button outlet: 'nextButton', class: 'btn', 'Find'
           @div class: 'btn-group btn-toggle btn-group-options', =>
-            @button outlet: 'regexOptionButton', class: 'btn', '.*'
-            @button outlet: 'caseOptionButton', class: 'btn', 'Aa'
-            @button outlet: 'selectionOptionButton', class: 'btn option-selection', '"'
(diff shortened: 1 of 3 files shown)


kevinsawicki:
> Does space pen have some way to do an "include" instead of @raw?

I don't think it does

simurai:
Ok, maybe good for now. Once we have more custom icons in other packages, we might can load them globally in a view.

Not sure why travis fails. If I run the package specs, it seems fine:

screen shot 2015-05-16 at 10 16 29 am

50Wliu:
@simurai restarted the build (it failed to download Atom correctly) and now it looks like everything is :green_heart:!

simurai:
@50Wliu Great. Thanks. Ok, I'll merge it.

PR #391, opened 2015-05-17
Add tooltip to whole word button

grizzilus:
There are currently tooltips for all buttons in the find view except the "whole word" option button. This PR adds the tooltip to that button to match the others in that view.


mnquintana:
Awesome, thanks for adding this! :ship:

PR #392, opened 2015-05-20
Fix accidental reassignment of path var

maxbrunsfeld:
Fixes a bug that caused exception to be thrown if you used the `Search in Directory` context menu command from the tree-view after already performing a project search.


Zireael07:
Thanks, I tripped that one a few times :)

PR #399, opened 2015-05-28
Make project search pluggable via DefaultSearchDirectoryProvider.

bolinfest:
This refactors `ResultsModel` so that when its `search()` method is
invoked, it finds the appropriate `SearchDirectoryProvider` for each
`Directory` in `atom.project.getDirectories()`. It invokes the `search()`
method for each provider and aggregates the results as they are streamed in.

Currently, there is only one implementation of `SearchDirectoryProvider`:
`DefaultSearchDirectoryProvider`. It uses the existing logic from
`ResultsModel` to implement `search()`, which is basically a call to
`atom.workspace.scan()` with an appropriate callback.

The next step is to make `find-and-replace` a consumer of `SearchDirectoryProvider`
objects so that remote directories can be supported. New providers will be added
to the front of the `ResultsModel::searchProviders` array so that they are considered
before the `DefaultSearchDirectoryProvider`. Note that the same technique is used
with `Project::directoryProviders`.

### DIFF
--- a/lib/project/default-search-directory-provider.coffee
+++ b/lib/project/default-search-directory-provider.coffee
@@ -1,0 +1,51 @@
+module.exports =
+class DefaultSearchDirectoryProvider
+  # Public: Determines whether this provider supports search for a `Directory`.
+  #
+  # * `directory` {Directory} whose search needs might be supported by this provider.
+  #
+  # Returns a `boolean` indicating whether this provider can search this `Directory`.
+  canSearchDirectory: (directory) -> true
+
+  # Public: Performs a text search for files in the specified `Directory`, subject to the
+  # specified parameters.
+  #
+  # Results are streamed back to the caller via `recordSearchResult()` and `recordSearchError()`.
+  #
+  # * `directory` {Directory} that has been accepted by this provider's `canSearchDirectory()`
+  # predicate.
+  # * `regex` {RegExp} to search with. (Note this reflects the "Use Regex" and "Match Case" options
+  # exposed via the ProjectFindView UI.)
+  # * `recordNumPathsSearched` {Function} callback that should be invoked periodically with the number of
+  # paths searched.
+  # * `recordSearchResult` {Function} Should be called with each matching search result.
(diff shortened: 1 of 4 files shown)


bolinfest:
This change depends on https://github.com/atom/atom/pull/6986.

Also, I'm hoping to get feedback on the API for `DefaultSearchDirectoryProvider` before investing the time to write tests for it. I think all of the methods are necessary: it's mainly a question of whether all of the parameters should be grouped into one object literal named `options`.

maxbrunsfeld:
@bolinfest It looks like your approach here and in atom/atom#6968 is to avoid using `Workspace::scan` for remote directories. What do you think about instead making this behavior (allowing custom search providers) *part of* `Workspace::scan`, so that this package could continue to use that method like it does currently?

In other words, extract the code inside of `Workspace::scan` that calls into `scandal` into a `DefaultDirectorySearcher` object that would live in Atom core.

maxbrunsfeld:
As a side note, it wasn't necessarily our intention that all atom services be named `{Something}Provider`. So don't feel too bound by that naming pattern. In this case, I might opt for something like `DirectorySearcher`. Just a thought.

bolinfest:
@maxbrunsfeld OK, that seems reasonable in this case: I'll give that a shot in a fresh PR.

The thing that I was trying to avoid, though it doesn't seem to be an issue in this case, is that the `iterator` would always have to run on the client. In general, this could be expensive if there are a ton of results and your iterator is acting as a filter. In that case, you would really like the filter to be applied on the server so that you don't have to send a ton of irrelevant of results over the wire. (Though that would mean the `iterator` has to be serializable, which is an issue.)

Though in this case, the `regex` argument is acting as that filter (which is serializable), and every item that is passed to the iterator is relevant, so it there is no issue with this use case.

bolinfest:
Take 2: https://github.com/atom/atom/pull/7022

bolinfest:
Closing in favor of https://github.com/atom/atom/pull/6986

PR #422, opened 2015-06-30
Prevent duplicate "Project Find Results" panels

### LINKED ISSUE #378: Open "Project Find Results" in Existing Tab Option
"Find in project" is always opening "Project Find Results tab in active Pane (or splitting active Pane if "Open Project Find Results in Right Pane" is enabled). I think the default behaviour (or at least an option) should be not to open new tab if "Project Find Results" tab already exists (as all 
"Project Find Results" tabs are updated anyway - new tab would make sens if old tabs would remain with previous results).

I would be very grateful for such a change/feature.

harai:
Prevent duplicate "Project Find Results" panels. Closes #378

### DIFF
--- a/lib/project-find-view.coffee
+++ b/lib/project-find-view.coffee
@@ -237,8 +237,8 @@ class ProjectFindView extends View
       @findEditor.getModel().selectAll()
 
   showResultPane: ->
-    options = null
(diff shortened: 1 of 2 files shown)


benogle:
Thanks for this, and thanks for blending in.

PR #426, opened 2015-07-01
createViews in activate

olofteskog:
The find-view has to exist, to be abel to use `find-and-replace:find-next-selected`, `find-and-replace:find-previous-selected` and `find-and-replace:use-selection-as-find-pattern` commands. This was not the case before this fix.

### DIFF
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -37,13 +37,11 @@ module.exports =
     @pathsHistory = new History(pathsHistory)
 
     @subscriptions.add atom.commands.add 'atom-workspace', 'project-find:show', =>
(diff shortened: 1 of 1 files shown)


benogle:
I'm not understanding what this fixes. Can you elaborate?

olofteskog:
Those commands are part of the find-view. If I use the commands, directly after launching atom, nothing will happen, because the find-view dont exist. So I have to toggle the findPanel to get them work.

maxbrunsfeld:
I'm going to close this out until you get a chance to address these issues. We can definitely reopen at a later time. Thanks.

PR #427, opened 2015-07-01
Scroll to first result on type

plrenaudin:
The idea of this pull request is to scroll the editor window on the result when you type, without having to press enter.
screencast from 01-07-15 21-46-01

### DIFF
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -17,6 +17,9 @@ module.exports =
     openProjectFindResultsInRightPane:
       type: 'boolean'
       default: false
+    scrollToResultOnType:
+      type: 'boolean'
+      default: false
 
   activate: ({@viewState, @projectViewState, @resultsModelState, @modelState, findHistory, replaceHistory, pathsHistory}={}) ->
     atom.workspace.addOpener (filePath) ->
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -17,7 +17,7 @@ module.exports =
     openProjectFindResultsInRightPane:
       type: 'boolean'
       default: false
-    scrollToResultOnType:
+    scrollToResultOnLiveSearch:
       type: 'boolean'
       default: false
 
--- a/spec/find-view-spec.coffee
+++ b/spec/find-view-spec.coffee
@@ -829,17 +829,25 @@ describe 'FindView', ->
       beforeEach ->
         findView.findEditor.focus()
 
-      it "scrolls to the first match depending on the settings", ->
-        editor.moveToTop()
+      it "scrolls to the first match if the settings scrollToResultOnLiveSearch is true", ->
         atom.config.set('find-and-replace.scrollToResultOnLiveSearch', true)
-        findView.findEditor.setText 'sort'
-        advance()
-        expect(editor.getSelectedBufferRange()).toEqual [[0, 9], [0, 13]]
-        expect(findView.findEditor).toHaveFocus()
+        editor.setHeight(3)
         editor.moveToTop()
-        atom.config.set('find-and-replace.scrollToResultOnLiveSearch', false)
-        findView.findEditor.setText 'concat'
+        originalScrollPosition = editor.getScrollTop()
+        findView.findEditor.setText 'Array'
         advance()
+        expect(editor.getScrollTop()).not.toEqual originalScrollPosition
+        expect(editor.getSelectedBufferRange()).toEqual [[11, 14], [11, 19]]
+        expect(findView.findEditor).toHaveFocus()
+
+      it "doesn't scroll to the first match if the settings scrollToResultOnLiveSearch is false", ->
+        atom.config.set('find-and-replace.scrollToResultOnLiveSearch', false)
+        editor.setHeight(3)
+        editor.moveToTop()
+        originalScrollPosition = editor.getScrollTop()
+        findView.findEditor.setText 'Array'
+        advance()
+        expect(editor.getScrollTop()).toEqual originalScrollPosition
         expect(editor.getSelectedBufferRange()).toEqual []
         expect(findView.findEditor).toHaveFocus()
 
--- a/spec/find-view-spec.coffee
+++ b/spec/find-view-spec.coffee
@@ -829,6 +829,20 @@ describe 'FindView', ->
       beforeEach ->
         findView.findEditor.focus()
 
+      it "scrolls to the first match depending on the settings", ->
+        editor.moveToTop()
+        atom.config.set('find-and-replace.scrollToResultOnLiveSearch', true)
+        findView.findEditor.setText 'sort'
+        advance()
+        expect(editor.getSelectedBufferRange()).toEqual [[0, 9], [0, 13]]
+        expect(findView.findEditor).toHaveFocus()
+        editor.moveToTop()
+        atom.config.set('find-and-replace.scrollToResultOnLiveSearch', false)
+        findView.findEditor.setText 'concat'
+        advance()
+        expect(editor.getSelectedBufferRange()).toEqual []
+        expect(findView.findEditor).toHaveFocus()
+
       it "updates the search results", ->
         expect(findView.descriptionLabel.text()).toContain "6 results"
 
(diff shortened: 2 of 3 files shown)


benogle:
I really really like the current behavior. I dont want to lose my place when typing in the find editor. If I want to move the editor, I can press enter.

plrenaudin:
This is another feature from many editors (Sublime Text again!) that I miss.
Maybe if it's controversial we could add this as a setting?

lib/find.coffee
```diff
@@ -17,6 +17,9 @@ module.exports =
     openProjectFindResultsInRightPane:
       type: 'boolean'
       default: false
+    scrollToResultOnType:
```
benogle: How about `scrollToResultOnLiveSearch`

benogle:
Can you add specs both with and without the setting enabled?

spec/find-view-spec.coffee
```diff
@@ -829,6 +829,20 @@ describe 'FindView', ->
       beforeEach ->
         findView.findEditor.focus()
 
+      it "scrolls to the first match depending on the settings", ->
```
benogle: Can you make this two specs? One with the setting on, and one with it off.

Also, this test doesnt look like it is testing that the scrollTop changes / doesnt change.

devios1337:
@benogle what possible reason can you have for *not* wanting to lose your place when typing in the find editor?

Ok let's play devil's advocate. Not wanting to lose your place in the editor would imply you are dealing with something in the current field of view. Since we know by necessity that you are also typing in a search term, we can only assume that the reason you don't want the editor to move while typing in a search term is that you are basing what you are typing on something that you are looking at in the current editor view.

We can only assume that this means you are typing something out that you are currently looking at. The alternative would be that you are looking at something and translating it on the fly into something else, and I motion that that scenario be considered a fringe case that would occur seldom at best.

*However*, firstly, this scenario should also be considered non-archetypal because it would be much more efficient for you to simply select the visible word and Cmd-E it, rather than type it all out again, and secondly, the editor is not likely to scroll away anyway since it will most likely be matching what you are already reading on the screen as you type it, unless the cursor is below what you are reading, in which case a simple Cmd-Shift-G will get you back there.

Even in the worst case scenario, where you end up getting completely lost in your document by typing the wrong thing, you can either backspace up or hit the Esc key to get back to your starting point with nothing lost.

In my opinion the advantages of a dynamically incrementing search far outweigh any disadvantages I can think of, but I'm curious to hear your reasoning.

thedaniel:
> In my opinion the advantages of a dynamically incrementing search far outweigh any disadvantages I can think of

Great, people that share your opinion can toggle this setting.

A project this large that has to account for so many conflicting opinions needs to be extremely conservative with behavior changes and defaults. Many significant behavior changes have followed and will follow the pattern of "add a setting but leave the current behavior the default" and "gather data and discuss changing the default in a major or minor release down the road". We're not picking on you or this concept (i will personally be toggling this setting because i prefer incremental search), it's just how things are done.

Looking forward to specs so this can merge.

plrenaudin:
Specs changed. Hope this will be merged soon.

benogle:
One reason is as daniel said, we're not going to change the default on folks until it's been out there for a while and it's clear the majority of people want it as the default.

> @benogle what possible reason can you have for not wanting to lose your place when typing in the find editor?

Let's say you want to search for `east`. You type `e`, and it moves your cursor to highlight the first `e`. There are a lot of `e`s in your doc, and it's almost certainly not what you want hightlighted. Then you type `a`, and it highlights `ea`, but there are a bunch of `easy`s in your doc. You dont care about the easy's, but now you're jumping a second time. Do you remember where you were? And on for the last 2 characters. Then you're non-committal, you just wanted to know where the `east` var name was used. But you cant go back to where you were. Nope.

With the current behavior you dont need to commit until you are ready for the jump. When you want to commit, you swing your pinky around and smack return, spending a total of .001 second.

Lets say you are in a function, you are picking a var name and you want to be consistent. What did you use in the other functions? Was it `packageName` or `packName`? `cmd-f`, type `packName`, notice there are 0 results. Ok, it must have been `packageName`. `esc`, keep typing your variable.

Or you're curious where you used the `packageName` variable name in this function. `cmd-f`, type `packName`, check out the highlights.

I'd say >50% of the time I open find, it's to do little checks like this. I'll keep track for a little while, just out of curiosity.

Does nearly everyone want one way or the other? We dont know. One could point to the issue indicating that a lot of people want this, but [history](https://github.com/atom/atom/issues/6605) tells us that the other side is not represented because it's already doing the things they want.

devios1337:
I also don't much like the idea of "people can't agree upon it, therefore make it a setting". Making something a setting is not a free solution. There is a cost of adding an option as a setting, not the least of which is complication of the application as a whole.

My reason for outlining example scenarios is to demonstrate usability. The purpose of a tool is to make people's jobs easier. It therefore completely matters how people use the tools and what their intentions and goals are in using a tool. I illustrated some counter-example scenarios with the explicit purpose of showing that they are not, in fact, good counter arguments. In fact I'm still waiting for a well-explained reason to keep the current behavior, even as an option. I'm not saying there aren't any, of course, just that I can't think of any strong reasons.

It does become harder to change functionality once people are dependent upon it, but in this case I argue that there is really not much that could be depended upon in the current implementation that would even be lost by adding this functionality. What, precisely, is the *advantage* of not changing the editor while the user is typing in a search term. What would actually be lost by adding this?

devios1337:
@benogle I still fail to see how your scenarios would be negatively impacted in any noticeable by changing this feature, aside from perhaps it being distracting to have the view change as you are typing. So you type "ea" and the first instance of "easy" is highlighted. Perhaps it's in a completely different part of the document and you are properly lost at this point.

You can still see, presumably quite obviously, that the selected word is actually "easy" and that's not what you are looking for. Immediately we have more information already than we did before, specifically that there is a word "easy" that comes before "east" in the document. We got this for free without having to explicitly ping the document by hitting Enter. It's not about how much time it takes to hit the enter key. I think most of us using this editor are pretty proficient at typing. It's about the mental effort—the requirement that an explicit action be taken that could have been automated. It is an extra action that is explicitly required and the user has to *remember* to type it.

This is the biggest problem I have personally with the current implementation: I keep forgetting to hit Enter. Granted this is probably because this is what I'm used to already working in other editors, but the argument I'm really trying to make here is that me forgetting to hit Enter and not getting the feedback I am expecting immediately is more detrimental than it is for you to have to hit Esc to get back to where you were, had it been incrementing in the first place. This, of course, is debatable.

However, even with your `packName` example, you'd still have to hit Esc to close the find box and put the keyboard focus back on the document to continue typing anyway, or use the mouse to click the document.

I'll give you that it is disorienting, and that may even be enough reason not to implement this, but I'm saying personally I don't see it as convincing.

dwelle:
@benogle 

1. your first use-case is quite a valid point but I'd argue __there should be a way to return to original caret position__ before you started searching. 
2. for the second use-case, your solution is sub-optimal. You should be using e.g. [highlight-selected][jrKgZ4] and select the word (via keyboard) instead of searching it.

[jrKgZ4]: https://atom.io/packages/highlight-selected

majelbstoat:
+1, looking forward to this support, and +1 that I should be able to Esc if I change my mind, and return to the original position.

tnrich:
+1 

Is this a feature yet?!

plrenaudin:
I'm not convinced by this: 
> I should be able to Esc if I change my mind, and return to the original position.
When people search, they are free to switch tab and let the find view opened. It would be weird to put them back to the position where they triggered the search then when they close the view.

barmstrong:
Agree, I think incremental search should be the default, just like in Sublime Text 2, Chrome, etc. There is a reason these apps have implemented it that way.

The main reason is that you often mis-type or mis-remember the exact name of things. When it finds while you type you are much more likely to find what you are looking for since it is giving you rapid feedback.

Simple example would be say I am searching for `really_long_file_name_of_something`

As I start typing it, the search starts to narrow. Maybe it finds it only after I've typed `really_lo` and I can early terminate typing the whole thing. Nice!

Another example, lets say I mis-remember it has underscores, and I start to type `really-lo...` with a dash instead. No results? Hmm...lets backup and try a few variations to narrow in on what I'm searching for.

Atom does this in the Cmd-T file selector and it works great. Same principle should apply to Cmd-F.

PR #430, opened 2015-07-02
Word and part of word selection logic

olofteskog:
I have a an suggestion on a Sublime-like select-next and select-all logic.

When using selecting the word under the cursor with select-next, it will match the next word as usual. But when manually marking the same word it will match next occurrence, even if it´s part of a word.
select-next

Same logic for select-all. 
select-all

### DIFF
--- a/lib/select-next.coffee
+++ b/lib/select-next.coffee
@@ -62,7 +62,10 @@ class SelectNext
     startingRange = options.start ? @editor.getSelectedBufferRange().end
     range = @findNextOccurrence([startingRange, @editor.getEofBufferPosition()])
     range ?= @findNextOccurrence([[0, 0], @editor.getSelections()[0].getBufferRange().start])
-    @addSelection(range) if range?
+    if range?
+      @addSelection(range)
+    else
+      @wordSelected = null
 
   findNextOccurrence: (scanRange) ->
     foundRange = null
--- a/lib/select-next.coffee
+++ b/lib/select-next.coffee
@@ -10,15 +10,23 @@ class SelectNext
 
   constructor: (@editor) ->
     @selectionRanges = []
+    @noofSelections = 0
 
   findAndSelectNext: ->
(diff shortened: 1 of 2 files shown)


benogle:
This is intentional. The whole words `for` that `cmd-d` is choosing are all the same, but the `for` in `fork` is not the same. It tries to match word boundaries of all matches so you dont accidentally over edit.

benogle:
Closing this as a wontfix.

shanemgrey:
I agree with the goal of not selecting parts of words when unwanted, but there are times when that's the desirable behavior.  

Example: I might copy this line and want to replace all instances of "disability" with "primaryLanguage"

``` ng-model="participant.disability" ng-options="disability.value as disability.text for disability in disabilityOptions" ```

``` ng-model="participant.primaryLanguage" ng-options="primaryLanguage.value as primaryLanguage.text for primaryLanguage in primaryLanguageOptions"```

The Find function has a "whole word" option which I think defaults to off, meaning Find will select **disability**Options while cmd-d and ctrl-cmd-g will not.  

Unfortunately, using Find and Replace All will select all instances in the buffer, when in this case I don't want to alter the original line, just the copy.  

With current functionality, I have the option of using Find and then putting in replacement text and using Replace for each next instance.  While this works, it's a bit more clunky than cmd-d to get multiple selections and typing over the selections.

I don't know the codebase well enough to understand what the fork from @olofteskog does exactly, but I think the desired behavior can be had by:

The hotkeys for 'find-and-replace:select-all' and 'find-and-replace:select-next' should both look to the current settings used by 'find-and-replace:show' and then follow the matching pattern in the same way.

This would allow more flexible functionality, keeping the settings all in one place, while maintaining whatever the desired functionality of the user is from one use to the next. 

I'm not experienced with coffeescript yet, but I'd be willing to work on this and submit a PR if you can point me at the necessary files to look at.

Issue #402, opened 2015-05-31
Uncaught TypeError: Cannot read property 'length' of undefined


walles:
[Enter steps to reproduce below:]

1. Double click on a word ('collect_highlights') in a Ruby file
2. Press cmd-g

Expected result:
* A search for occurrences of 'collect_highlights'

Current result:
* An error dialog

**Atom Version**: 0.204.0
**System**: Mac OS X 10.10.3
**Thrown From**: [find-and-replace](https://github.com/atom/find-and-replace) package, v0.169.0

### Stack Trace

Uncaught TypeError: Cannot read property 'length' of undefined

```
At /Users/johan/Downloads/Atom.app/Contents/Resources/app.asar/node_modules/find-and-replace/lib/find-view.js:449

TypeError: Cannot read property 'length' of undefined
  at FindView.module.exports.FindView.findAndSelectResult (/Users/johan/Downloads/Atom.app/Contents/Resources/app.asar/node_modules/find-and-replace/lib/find-view.js:449:23)
  at FindView.findAndSelectResult (/Users/johan/Downloads/Atom.app/Contents/Resources/app.asar/node_modules/find-and-replace/lib/find-view.js:3:61)
  at FindView.module.exports.FindView.findNext (/Users/johan/Downloads/Atom.app/Contents/Resources/app.asar/node_modules/find-and-replace/lib/find-view.js:429:19)
  at FindView.findNext (/Users/johan/Downloads/Atom.app/Contents/Resources/app.asar/node_modules/find-and-replace/lib/find-view.js:3:61)
  at atom-workspace.subscriptions.add.atom.commands.add.find-and-replace:find-next (/Users/johan/Downloads/Atom.app/Contents/Resources/app.asar/node_modules/find-and-replace/lib/find-view.js:343:26)
  at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/Users/johan/Downloads/Atom.app/Contents/Resources/app.asar/src/command-registry.js:238:29)
  at /Users/johan/Downloads/Atom.app/Contents/Resources/app.asar/src/command-registry.js:3:61
  at KeymapManager.module.exports.KeymapManager.dispatchCommandEvent (/Users/johan/Downloads/Atom.app/Contents/Resources/app.asar/node_modules/atom-keymap/lib/keymap-manager.js:520:16)
[...]

PR #434, opened 2015-07-05
Don’t throw exception when replacing with an empty find string

izuzak:
This fixes https://github.com/atom/find-and-replace/issues/402#issuecomment-118548573 and closes https://github.com/atom/find-and-replace/issues/402 (since the initial report was fixed in ed00ffb, I believe):

* 466c592 adds a [failing spec](https://travis-ci.org/atom/find-and-replace/builds/69598472) to demonstrate the problem
* a519074 [fixes the problem](https://travis-ci.org/atom/find-and-replace/builds/69598769) by checking that `@markers` is defined before calling `.length` on it.

cc @kevinsawicki @benogle for :eyes:

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -258,15 +258,15 @@ class FindView extends View
     @findHistory.store()
     @replaceHistory.store()
 
-    if @markers.length is 0
-      atom.beep()
-    else
+    if @markers?.length > 0
       unless currentMarker = @findModel.currentResultMarker
         markerIndex = @[nextIndexFn]()
         currentMarker = @markers[markerIndex]
 
       @findModel.replace([currentMarker], @replaceEditor.getText())
       @[nextOrPreviousFn](fieldToFocus: @replaceEditor)
+    else
+      atom.beep()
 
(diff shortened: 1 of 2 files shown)


lee-dohm:
:+1: Looks good to me

Additionally, rather than simply beeping, should we display an auto-hiding warning notification about what the problem is? I get really frustrated when I try to do something and whatever program just beeps at me. I feel like it's roughly the equivalent of the dialog box: "An error occurred. Ok?"

izuzak:
> Additionally, rather than simply beeping, should we display an auto-hiding warning notification about what the problem is? I get really frustrated when I try to do something and whatever program just beeps at me. I feel like it's roughly the equivalent of the dialog box: "An error occurred. Ok?"

I definitely wouldn't mind discussing that in a new issue :+1: since that pattern is used in a few places and  also keeping one PR to one thing has historically been a Good Decision :tm:. :)

lee-dohm:
You're absolutely right. Done :grinning:

benogle:
:+1:

Issue #339, opened 2015-01-26
Provide find by whole words option for project find.


jeffschwartz:
Currently this isn't implemented in project find.

therealmarv:
+1 ... it was causing headaches for me today.
A tip from @timscott about this issue in file search for a workaround for now:

Use regex and search for `\bword\b`

Zentrifuge:
+1

opadillaintegrify:
Looks like we came full circle.  This originally started as a way to do a whole word search *without* having to use regex

https://discuss.atom.io/t/how-to-search-for-whole-words-in-atom/10969/9

which lead to 
https://github.com/atom/find-and-replace/pull/260

which lead to this issue

changnet:
so far,use regex,and search \bfoo\b instedad.foo is the whole word you search

PR #435, opened 2015-07-05
Add whole word option for project find

kubetz:
This PR closes https://github.com/atom/find-and-replace/issues/339 .

whole-word
whole-word-tooltip

Tests were heavily inspired by the "case sensitivity" tests.

Looking forward to your feedback.


benogle:
Thanks so much for blending in :heart:

Issue #331, opened 2014-12-18
Add undo to replace all in project


hachpai:
Hello,

I've spent 2 hours to undo a bad replace all in project. An not-undoable operation for the moment.
It's REALLY dangerous :-1: 

Miss this function truly.

thank you

ph

samtuke:
Yeah this is crazy. I just experienced:

> Replaced " with '  41197 times in 961 files

Why does Atom replace in files which aren't open? Why did it replace in multiple files in the first place? I see only a "replace all" button, I'm unsure if the changes were saved across all affected files already (appears so, no blue icons on tabs of open files indicating unsaved changes).

Ouch :(

benogle:
> Why did it replace in multiple files in the first place?

There are two kinds of find and replace UIs: find in buffer, and find in project. Sound like find in buffer was what you wanted, but you used find in project.

JacobEvelyn:
+1, I've run into this several times (including now) and it's a huge pain to fix the mistake.

dk253:
twice this week... the last one was 1350 changes in 58 files.  Woohoo.

samtuke:
On 03/03/15 15:33, dk253 wrote:
> twice this week... the last one was 1350 changes in 58 files. Woohoo.

Heck. I feel your pain.

btm1:
i just had this happen too... also where is there only a replace all when you find in project? I just want to find all instances of something. I could have sworn there used to be just a find button.

waywalker:
Wow. Danger :mouse: :-) Just got bitten by this as well.

park9140:
Perhaps a confirm when replacing in unopened files would be appropriate until we can get undo?

andya9:
:+1:  it would be really really useful

graemeworthy:
+1

jb:
I just did this on my entire `~/.atom` directory when trying to replace something in package file. If I'd accidentally opened my home directory instead (which is quite possible for a newb to do by accident), I likely would have lost data.

IMO, this feature should be removed entirely until some safeguards can be put in place, such as:

  - A warning that all files in directory X will be modified
  - A warning that appears if some of the files to be modified are not version-controlled
  - A warning that appears if the project directory is `/`, `/home/*/`, or `/User/*/`
  - An undo feature (that will still work if Atom crashes during the replace)

[...]

PR #446, opened 2015-07-13
Disable hotkey for `project-find:show`

zaz:
Bearing in mind that many people have reported losing hours of work to bug #331, and that those people are only a subset of the people affected, it seems likely that hours of work are being lost per day.

While this solution is not ideal, replacing in the whole project is not a common operation, so removing its hotkey will likely cause less time to be wasted than if we kept the hotkey.

I know this may seem like a poor solution, but please don't reject this PR unless you really believe that the hotkey is saving more time than it's costing, or if you're going to add a better solution very soon.

Thanks.

### DIFF
--- a/keymaps/find-and-replace.cson
+++ b/keymaps/find-and-replace.cson
@@ -1,10 +1,10 @@
 '.platform-darwin':
-  'cmd-F': 'project-find:show'
+  # 'cmd-F': 'project-find:show'
   'cmd-f': 'find-and-replace:show'
   'cmd-alt-f': 'find-and-replace:show-replace'
 
 '.platform-win32, .platform-linux':
-  'ctrl-F': 'project-find:show'
+  # 'ctrl-F': 'project-find:show'
   'ctrl-f': 'find-and-replace:show'
   'ctrl-alt-f': 'find-and-replace:show-replace'
 


benogle:
Sorry, this is not the correct solution to this. A better solution would be a confirm box. Something like "Are you sure you want to replace 'before' with 'after' 100 times in 34 files?"

zaz:
I trust you to make the right decision on this, but I have one last thing to say:

You're right, but I really think that removing this feature entirely right now would make Atom a better product. I think your experience with Atom is possibly skewing your perspective as to how bad the issue actually is. People assume that Atom, being a well-designed tool like Git, will not erase data unless instructed to do so explicitly. Every other action I can remember in Atom has been undoable, but here that safety net has suddenly been taken away without warning, breaking the user's trust.

Thanks for the work you're doing, and sorry if I wasted your time.

50Wliu:
> You're right, but I really think that removing this feature entirely right now would make Atom a better product.

The problem with this is that if it's removed (even temporarily for a release or two!) the issue trackers will be flooded with reports of Project Find not working.  If I had to guess, many more people use Find in Project then those who hit #331.  That's not to say it's a minor issue (it definitely is super frustrating), but I also think that the benefits of removing the keybinding doesn't outweigh the consequences.

zaz:
Okay, that's fair enough. Sorry for being unhelpful.

PR #456, opened 2015-07-22
Escape search on use-selection-as-find-pattern and Regex enabled #455

### LINKED ISSUE #455: Escape search string on use-selection-as-find-pattern and Regex enabled
Currently, if the Regex option is enabled and the `use-selection-as-find-pattern` command is triggered on a string like `function foo($bar) {`, then the find-window's search string won't actually match the selection since the characters in the string don't make for a valid regular expression.

If the Regex option is selected, then the selection string should be escaped when triggering `use-selection-as-find-pattern` so that it is a valid regular expression.

adamfranco:
Currently, if the Regex option is enabled and the use-selection-as-find-pattern command is triggered on a string like `function foo($bar) {`, then the find-window's search string won't actually match the selection since the characters in the string don't make for a valid regular expression.

If the Regex option is selected, then the selection string should be escaped so that it is a valid regular expression.

This patch escapes the search string when the regex option is enabled, allowing the resulting pattern to match the selection.

Closes #455

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -197,8 +197,7 @@ class FindView extends View
   focusFindEditor: =>
     selectedText = atom.workspace.getActiveTextEditor()?.getSelectedText?()
     if selectedText and selectedText.indexOf('\n') < 0
-      if @findModel.useRegex
-        selectedText = Util.escapeRegex(selectedText)
+      selectedText = Util.escapeRegex(selectedText) if @findModel.useRegex
       @findEditor.setText(selectedText)
     @findEditor.focus()
     @findEditor.getModel().selectAll()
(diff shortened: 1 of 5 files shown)


lib/find-view.coffee
```diff
@@ -196,6 +197,8 @@ class FindView extends View
   focusFindEditor: =>
     selectedText = atom.workspace.getActiveTextEditor()?.getSelectedText?()
     if selectedText and selectedText.indexOf('\n') < 0
+      if @findModel.useRegex
```
benogle: Minor, but can you use postfix ifs on these?

lib/find-view.coffee
```diff
@@ -379,6 +382,8 @@ class FindView extends View
     editor = @findModel.getEditor()
     if editor?
       pattern = editor.getSelectedText() or editor.getWordUnderCursor()
+      if @findModel.useRegex
```
benogle: Same: postfix if here

benogle:
This looks good other than the formatting changes

adamfranco:
Thanks for reviewing @benogle , I've updated the code-formatting as suggested.

benogle:
Thanks for this!

PR #471, opened 2015-08-03
Update project replace-all flow 

benogle:
This is a tuneup for https://github.com/atom/find-and-replace/issues/331. People are hosing their working state and it sucks. Replace all was a little too eager. This PR changes a few things

First, you must have run a search with results to enable the replace all functionality. Before a search or after a search with no results, the replace-all function is disabled, and the command will 'beep':

disasbled

The second thing is a confirm box on replace all:

screen shot 2015-08-03 at 4 33 09 pm

And the third is that it will not replace unless the search pattern is the same as the results pattern. Previously, if you searched for `welcome`, changed the find box to `something`, then ran a replace all, it would first search for `something`, then do the replace all on `something`. Now you must explicitly search for the string before you can replace it:

fnr-replace-all

Hopefully these checks will make things more reliable for folks.

### DIFF
--- a/lib/project-find-view.coffee
+++ b/lib/project-find-view.coffee
@@ -147,6 +147,8 @@ class ProjectFindView extends View
     @subscriptions.add new Disposable ->
       $(window).off 'focus', focusCallback
 
+    @findEditor.getModel().getBuffer().onDidChange =>
+      @updateReplaceAllButtonEnablement(@model.getResultsSummary())
     @handleEventsForReplace()
 
   handleEventsForReplace: ->
@@ -231,6 +233,14 @@ class ProjectFindView extends View
   replaceAll: ->
     return atom.beep() unless @model.matchCount
 
+    currentPattern = @findEditor.getText()
+    if @model.pattern isnt currentPattern
+      atom.confirm
+        message: "The searched pattern '#{@model.pattern}' was changed to '#{currentPattern}'"
+        detailedMessage: "Please run the search with the new pattern '#{currentPattern}' before running a replace-all"
+        buttons: ['OK']
+      return
+
     @showResultPane().then =>
       pattern = @model.pattern
       replacementPattern = @replaceEditor.getText()
@@ -239,7 +249,7 @@ class ProjectFindView extends View
       buttonChosen = atom.confirm
         message: 'Are you sure you want to replace all?'
         detailedMessage: message
-        buttons: ['Ok', 'Cancel']
+        buttons: ['OK', 'Cancel']
 
       if buttonChosen is 0
         @clearMessages()
@@ -297,7 +307,8 @@ class ProjectFindView extends View
     @descriptionLabel.html(errorMessage).addClass('text-error')
 
   updateReplaceAllButtonEnablement: (results) ->
-    @replaceAllButton[0].disabled = !results?.matchCount
+    canReplace = results?.matchCount and results?.pattern is @findEditor.getText()
+    @replaceAllButton[0].disabled = !canReplace
 
   updateOptionsLabel: ->
     label = []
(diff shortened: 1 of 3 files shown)


lee-dohm:
I've thought of two alternates for the "change the search string after the search and then click Replace All" workflow ...

1. Disable the Replace All button until they search again
1. Don't disable the Replace All button ... but when they click it, make it search the first time they click it and then the second click replaces

benogle:
> Disable the Replace All button until they search again

Ok, this is what it does now. See the OP for a gif. It seems super paranoid, but I think it might be a good thing.

lee-dohm:
Looks good to me! :rocket:

thomasjo:
I say we :ship: this!

bronson:
I like this a lot.

Is it obvious why the button is grayed out?  (I can't tell  from the gifs)  It's super frustrating when there's a button or menu item that I need, and there's no indication why it's grayed out.

mnquintana:
> Is it obvious why the button is grayed out? (I can't tell from the gifs) It's super frustrating when there's a button or menu item that I need, and there's no indication why it's grayed out.

Hmm that's a good point - might be worth adding an explanatory tooltip when Replace All is disabled.

benogle:
Currently tooltips dont show on disabled buttons. Hopefully it's straightforward enough that there should be a search before a replace. I'll merge this, and if folks are confused, we can dig into the tooltip issue.

brianmhunt:
This is a great step, good work.

Making the replacements undoable would be even better for people with slippery fingers :grin:.

benogle:
> Making the replacements undoable would be even better for people with slippery fingers .

Please see this comment https://github.com/atom/find-and-replace/issues/331#issuecomment-126833569

brianmhunt:
Thanks @benogle 

There are a few constraints, but the mass-replace undo is not insurmountable. The simplest solution is:

1. Store every change (file/line/column/was/becomes) in a "transaction"
2. When the user makes any change† after the replace-all, the transaction becomes stale and unusable for undo.
3. If the user does undo, it's a trivial find and replace.

† or deviates from the state of any file changed by the replace-all, to which state they can return by using the local undo.

The memory and processing requirements are `O(n)` for `n` changes, so practical for all but massive numbers of changes.

There are more elaborate alternatives, such as keeping track of the set of changes in a separate buffer and making individual changes undoable in a tab one can open to review them.

All to say, an undo should be quite doable, and from a UX perspective it's preferable to IMHO a confirmation dialog.

bronson:
@brianmhunt I don't understand... undoing a find-and-replace is itself a trivial find-and-replace?  How would that be?

brianmhunt:
@bronson in oversimplified pseudocode 

```javascript
for (tr in transactions) {
  select(tr.file, tr.line, tr.column, tr.replacement.length)
    .replaceWith(tr.was)
}
```

bronson:
@brianmhunt sure, before you throw away the saved data.  What about after the user makes a change?

(Atom's idea of a transaction is similar enough that I hesitate to use the term: https://atom.io/docs/api/v1.0.2/TextBuffer#instance-transact)

brianmhunt:
@bronson Yes, good point on being careful not to conflate the "transaction" term with Atom internals.

As in point #2 above, at its most basic once the user makes a change the transaction history would be made stale and unusable.

bronson:
Right.  So what does point #3 mean?

Or, are you saying that the moment the user makes a change after a find-and-replace, all previous undos are suddenly lost?  That doesn't sound so hot to me...?

brianmhunt:
@bronson Point #3 is the conditional operation that executes the "replace-all-undo" should the user elect to do so. This is not to be confused with the regular "undo" stack.

So the Replace-All function would have its own separate `undo` concept that is separate from the regular undo stack. This simplifies the development scope substantially; adding the "replace-all-undo" into the existing undo stack would be somewhat more complex, and very possibly impractical.

thomasjo:
PRs are welcome :smile:

PR #474, opened 2015-08-05
Regex syntax highlighting

### LINKED ISSUE #458: Add syntax highlighting in Regex mode
Highlighting syntax in Regex mode would be a nice usability enhancement. This is something the BBEdit editor does very well, highlighting regular expressing special characters in one color and escaped characters in another:

bbedit-find-highlight-a

Doing this makes it much easier to construct and modify regex search strings as code often contains  many characters that are also special regular-expression syntax characters. Additionally, highlighting placeholders and escaped characters in the replacement string would make them stand out more from the static replacement text.

adamfranco:
This PR implements syntax highlighting for regular expression searches and will fix #458.

Highlighting works in both the buffer-find and project find dialogs and defaults to text colors that are required by all themes. Themes can override the styling to tweak colors as they desire.

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -296,6 +296,12 @@ class FindView extends View
 
   updateModel: (options) ->
     @findModel.setSearchParams(options)
+    # Set syntax highlighting if using regular expressions
+    if @findModel.useRegex
+      grammar = atom.grammars.grammarForScopeName('source.regexp.python')
+      @findEditor.getModel().setGrammar(grammar)
+    else
+      @findEditor.getModel().setGrammar(atom.grammars.nullGrammar)
 
   updateResultCounter: =>
     if @findModel.currentResultMarker and (index = @markers.indexOf(@findModel.currentResultMarker)) > -1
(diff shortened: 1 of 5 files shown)


benogle:
Can you add some specs to this? You can test that the grammar is added and removed when the regex setting is toggled in both the find-view and the project-find-view.

benogle:
Also, thanks for sticking with this. It will be a good addition a lot of people will like.

adamfranco:
I hadn't figured out what specs would be useful, thanks for the tip. I'll add these when I'm back at a computer this weekend. Cheers!

adamfranco:
I've now added specs that test that the grammars are getting set and unset properly when the regex toggle buttons are pressed or regex mode was set in the config. It took me a while to figure out that I needed to specifically enable the language-javascript package in the spec to ensure its grammars are available, but it's working now. :smile: 

As far as I can tell, this PR should be ready to go.

benogle:
Nicely distilled. :bow:

benogle:
I assume highlighting of replacements will be in another PR?

benogle:
Hey this is on twitter :p https://twitter.com/atomeditor/status/631947977527955456

adamfranco:
Thanks for the Twitter link @benogle , glad folks are enjoying it. :grinning: 

By the way, the regex replacement grammar in atom/language-javascript#198 has just been merged into language-javascript, so #494 (adding highlighting of replacement syntax to the find-and-replace dialog) should begin passing tests shortly, closing out this set of enhancements.

PR #477, opened 2015-08-07
Share buffers and history cyclers between buffer and project panels

benogle:
This is an extension of #475, and makes it so the find box in both panels feels exactly the same. 

A manic gif that probably doesnt make sense:

shared-cycler

TODO

* [x] Make it work
* [x] Add the editor extension to atom-space-pen-views https://github.com/atom/atom-space-pen-views/pull/15
* [x] Specs


lib/text-editor-view.coffee
```diff
@@ -0,0 +1,47 @@
...
+class TextEditorView extends View
+  # The constructor for setting up an `TextEditorView` instance.
+  constructor: (params={}) ->
+    {mini, placeholderText, attributes, editor} = params
+    attributes ?= {}
+    attributes['mini'] = mini if mini?
+    attributes['placeholder-text'] = placeholderText if placeholderText?
+
+    if editor?
+      @element = atom.views.getView(editor)
```
benogle: Duplicating this file is no bueno. TODO: update atom-space-pen-views to allow for the `editor` param.

lib/find.coffee
```diff
@@ -135,10 +135,18 @@ module.exports =
   createViews: ->
     return if @findView?
 
-    history = {@findHistory, @replaceHistory, @pathsHistory}
+    findBuffer = new TextEditor(mini: true).getBuffer()
```
nathansobo: I'm not fully understanding what's going on here... looks like you're creating these editors just to get access to their buffers? Why not just get access to the buffer of the editors you actually create?

benogle: Yeah. I wanted them to be consistent in the way they created the editors between the views. Looks like `TextBuffer` is exported. I can just create them directly.

nathansobo: :+1:

PR #482, opened 2015-08-11
Use wholeWord option in select-next (cmd-d)

### LINKED ISSUE #340: Select next misses matches when starting with selection at the end of a word

### LINKED ISSUE #476: CMD-D Failing to Capture Match
If I have the following code, and select the first instance of "update" and press `cmd+d` to get the other two, only the third is selected.

atom

However, if I select the second instance of "update" and press `cmd+d` it will select the third then the first instance.

atom2

I have `highlight-selected` installed and you see that it finds the second instance (in the first scenario) just fine.

benogle:
Currently when using `cmd-d`, we're checking if the selected text is a word, if so, we only select words, and skip matches within other words. Tons of people think this is a bug.

Someone else's gif: 

This PR retains the current behavior when the `wholeWord` option is enabled in one of the find panels, and will match all the things when `wholeWord` is false.

Closes atom/atom#6800
Closes #476
Closes #340
Closes #430
Closes #356
Closes #450

### DIFF
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -118,7 +118,7 @@ module.exports =
       editor = editorElement.getModel()
       selectNext = @selectNextObjects.get(editor)
       unless selectNext?
-        selectNext = new SelectNext(editor)
+        selectNext = new SelectNext(editor, {@findOptions})
         @selectNextObjects.set(editor, selectNext)
       selectNext
 
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -48,7 +48,7 @@ module.exports =
       else
         @findModel.setEditor(null)
 
-    @subscriptions.add atom.commands.add '.find-and-replace, .project-find', 'window:focus-next-pane', =>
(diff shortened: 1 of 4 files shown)


jhasse:
Thank you so much for this! :)

maxbrunsfeld:
Woah just got the hang of this. Pretty nice.

colllin:
In my opinion, the real bug here was that the find palette is still highlighting things even after you've conceptually departed from the find palette.  In this example, is it considered a bug that `cmd-d` doesn't select `crowdTypes`, as highlighted by the find-and-replace search?

cmd-d is unrelated to find-and-replace

I'm working on a proposal / PR for how to reconcile these entangled and conflicting behaviors.

PR #489, opened 2015-08-12
Force search results outlines to be in front of other highlights

### LINKED ISSUE #406: The border of the selected result is missing
Only if I previously selected something then the current find result has border, otherwise there will be mistakenly no border. I don't mean the borders of all results, but just the border of the current result. For example in "One Light" theme the borders of all results are black and the border of selected result should be blue.

as-cii:
Fixes #406.

return

/cc: @benogle @izuzak


benogle:
Sweet thanks!

Issue #297, opened 2014-10-08
No Find button on Find In Project panel


CharlesJenkins:
As shown in this project's readme, the Find In Project panel has a Replace All button, but no Find button. If the user doesn't want to replace, there seems to be no way to find. The user has no way to know it's safe to press Enter to execute the Find operation. Please consider adding a Find button to provide an obviously safe way to do a search without replacing.

splendido:
+1 for adding the button!

thisisnian:
Could this be reopened? It would be nice to be on a todo list.
I can't think of a reason why it can't be left open and looked at some point in the future,

Happy to look at it myself if the issue is reopened.

izuzak:
I think this is worth reconsidering -- I've reopened and added an "enhancement" label.

benogle:
This was added in #509. Thanks @thisisnian!

hoangduytranuk:
I hope someone is reading this, and on the subject of adding extra buttons in search panel, I would like to have two arrowed buttons, indicating search back (arrow up)/forth (arrow down). I don't think the code will be much as functionality is ALREADY there, just adding two graphical buttons and a couple of arrow icons.

PR #509, opened 2015-08-25
Project View Search Button

thisisnian:
This is to create a 'Find' button so that users are able to press Find instead of having to press enter.
This is to improve user friendiness, instead of having to guess enter = find.

This is my first dabble in coffee, code review appreciated. 

Please see https://github.com/atom/find-and-replace/issues/297

Cheers,

----

Added by @simurai

Before:

screen shot 2015-09-22 at 10 48 29 pm

After:

screen shot 2015-09-22 at 10 46 07 pm

Once this gets merged, also merge One dark/light PRs https://github.com/atom/one-dark-ui/issues/91 + https://github.com/atom/one-light-ui/issues/35.

### DIFF
--- a/lib/project-find-view.coffee
+++ b/lib/project-find-view.coffee
@@ -35,7 +35,7 @@ class ProjectFindView extends View
       buffer: pathsBuffer
       placeholderText: 'File/directory pattern. eg. `src` to search in the "src" directory or `*.js` to search all javascript files.'
 
-    @div tabIndex: -1, class: 'project-find padded', =>
+    @div tabIndex: -1, class: 'find-and-replace', =>
       @header class: 'header', =>
         @span outlet: 'descriptionLabel', class: 'header-item description'
         @span class: 'header-item options-label pull-right', =>
@@ -52,7 +52,7 @@ class ProjectFindView extends View
         @div class: 'input-block-item', =>
           @div class: 'btn-group btn-group-find', =>
             @button outlet: 'findAllButton', class: 'btn', 'Find'
-        @div class: 'input-block-item', =>
+
           @div class: 'btn-group btn-toggle btn-group-options', =>
             @button outlet: 'regexOptionButton', class: 'btn option-regex', =>
               @raw '<svg class="icon"><use xlink:href="#find-and-replace-icon-regex" /></svg>'
--- a/lib/project-find-view.coffee
+++ b/lib/project-find-view.coffee
@@ -52,7 +52,6 @@ class ProjectFindView extends View
         @div class: 'input-block-item', =>
           @div class: 'btn-group btn-group-find', =>
             @button outlet: 'findAllButton', class: 'btn', 'Find'
-
           @div class: 'btn-group btn-toggle btn-group-options', =>
             @button outlet: 'regexOptionButton', class: 'btn option-regex', =>
               @raw '<svg class="icon"><use xlink:href="#find-and-replace-icon-regex" /></svg>'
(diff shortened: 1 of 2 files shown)


mertkahyaoglu:
IMO, just pressing enter is fast and very user friendly :grin: but let's see anyone want this. The only thing bothers me about the code is additional line breaks.

thisisnian:
Well there have been some users (myself included) that wasn't sure how to search, either this or even an 'Enter to Search' tooltip or something. 

In any case, users can still press the enter button, whatever floats your boat.

I've just fixed the formatting.

lib/project-find-view.coffee
```diff
@@ -365,4 +374,4 @@ class ProjectFindView extends View
     @search(onlyRunIfActive: true, caseSensitive: not @model.getFindOptions().caseSensitive)
 
   toggleWholeWordOption: ->
-    @search(onlyRunIfActive: true, wholeWord: not @model.getFindOptions().wholeWord)
+    @search(onlyRunIfActive: true, wholeWord: not @model.getFindOptions().wholeWord)
```
50Wliu: This space isn't needed and also causes the linter tests to fail.

thisisnian:
thanks must have missed that one.

simurai:
How about changing the button widths a bit so that the inputs line up? Something like:

screen shot 2015-08-26 at 4 50 19 pm

I can help out with the styling if you like it.

thisisnian:
@simurai Got it like the screenshot above, and I've got it built in Linux. However it fails Travis CI. I guess I'm doing something wrong?

simurai:
I think the check fails because `project-find` got renamed to `find-and-replace`.

thisisnian:
I reverted the failing changes on my local instance, this is what I got 

image

I'm not pretty sure where the property (boxed in red) comes from. Looks like it may be from another package as the one defined in the styles gets overwritten. Has to be more than 10em otherwise the buttons are too small. 

Gah I feel so stupid, must be something really simple that I'm missing.

simurai:
> I'm not pretty sure where the property (boxed in red) comes from. Looks like it may be from another package as the one defined in the styles gets overwritten.

Yes, right. That comes from the One dark theme. If you switch to another theme (like Atom dark), then it should look fine. I'll make another PR for the One dark theme to fix it. It needs the `em`s so that font sizing can be changed.

__Tip__: In case you'd like to see where the styles come from, you can click the `<style>...</style>`, then it should highlight the less file.

screen shot 2015-08-29 at 12 20 12 pm
screen shot 2015-08-29 at 12 20 44 pm

thisisnian:
Ah I knew it came from there, however I was just unsure as it would mean potentially not just one-dark theme that needs changing, however anyone with a custom theme that didn't have the correct css attributes (could be many in the future)

Thanks for checking over it.

simurai:
> I was just unsure as it would mean potentially not just one-dark theme that needs changing, however anyone with a custom theme that didn't have the correct css attributes (could be many in the future)

Typically, themes don't have to bother about styling those buttons. Unless they wanna customize it. The One themes override some of the `px` values to make it possible to change the font size so the whole UI can be resized. But that shouldn't be the concern of this package.

It *is* a bit of a problem when people fork the One themes, since those themes would need a fix too.

simurai:
@benogle How do you like this change of adding a `Find` button to project search? I quite like it:

- Better discovery for beginners
- More consistent with the find in buffer.

benogle:
I think it makes sense. It would be nice if there were a test. Other than than, the code looks fine.

PR #536, opened 2015-09-24
Adapt specs to work with the logical DisplayBuffer

as-cii:
This changes the specs so that they use the editor element to test visual changes, so that they still continue to pass as soon as we merge atom/atom#8905.

/cc: @nathansobo


nathansobo:
Same here... is this mandatory? We prefer to always keep package specs passing against the current release.

PR #541, opened 2015-09-30
Use findOptions.caseSensitive when selecting next occurrence (cmd-d)

jeancroy:
Address https://github.com/atom/atom/issues/9003

This was a fun one liner... until it was time to write some specs - there's tons of options permutations !

Ultimately the various cases will use one of 2 regex with 2 flag options for each, so that is 4 tests.
( Note that 2 of those 4 cases are the normal/current behavior so there would be no harm removing them. )

Case sensitive was the default everywhere before this PR, so I explicitly set it to true on other tests to keep things as they where.


benogle:
Thanks :+1:

PR #546, opened 2015-10-05
Add find in parent directory functionality

codingbelief:
This request is to add finding the selected word in the parent directory functionality and mapping to 'alt-a' key.
Finding in parent directory is frequently used when reading souce code.
And using hot key is more convenient than using the 'Search In Directory' context menu command.

### DIFF
--- a/keymaps/find-and-replace.cson
+++ b/keymaps/find-and-replace.cson
@@ -4,7 +4,7 @@
   'cmd-alt-f': 'find-and-replace:show-replace'
 
 'atom-text-editor':
-  'alt-a': 'project-find:show-in-parent-directory'
+  'alt-a': 'project-find:show-in-current-directory'
 
 '.platform-win32, .platform-linux':
   'ctrl-F': 'project-find:show'
--- a/keymaps/find-and-replace.cson
+++ b/keymaps/find-and-replace.cson
@@ -3,6 +3,9 @@
   'cmd-f': 'find-and-replace:show'
   'cmd-alt-f': 'find-and-replace:show-replace'
 
+'atom-text-editor':
+  'alt-a': 'project-find:show-in-parent-directory'
+
 '.platform-win32, .platform-linux':
   'ctrl-F': 'project-find:show'
   'ctrl-f': 'find-and-replace:show'
--- a/README.md
+++ b/README.md
@@ -9,3 +9,7 @@ Find and replace in the current buffer or across the entire project in Atom.
 ## Find in project (cmd-shift-f)
 
 ![screen shot 2013-11-26 at 12 26 02 pm](https://f.cloud.github.com/assets/69169/1625945/b216d7b8-56d9-11e3-8b14-6afc33467be9.png)
+
+## Provided Service
+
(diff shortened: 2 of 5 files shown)


keymaps/find-and-replace.cson
```diff
@@ -3,6 +3,9 @@
   'cmd-f': 'find-and-replace:show'
   'cmd-alt-f': 'find-and-replace:show-replace'
 
+'atom-text-editor':
+  'alt-a': 'project-find:show-in-parent-directory'
```
benogle: Why `alt-a` Is it used in other editors?

benogle:
This seems like the same functionality as the `show-in-current-directory`. Can you make the codepaths the same and use `show-in-current-directory` for the keybinding?

benogle:
Also there is a linter error, and this does not have specs.

keymaps/find-and-replace.cson
codingbelief: I'm not finding any other editors using `alt-a`, I just can't find other free keybinding of two keys which can be handled by one hand. 

I try to use `alt-f` at first, but it is used as 'editor:move-to-end-of-word' by atom core.

timkelty:
This works for me, but I have to do the keymap twice - once brings up find (empty), a second populates the current dir.

timkelty:
@codingbelief I believe having to do the keymap twice is a result of having the following setting:
```
  tabs:
    usePreviewTabs: true
```

If the current file is fully open (you have double clicked it from tree-view or edited it), then `show-in-current-directory` works right away. However, if you are previewing the file, you have to do `show-in-current-directory` twice to get the path in there.

timkelty:
Is there anything that can be done to move this along?
I've been running @codingbelief's fork for a couple weeks now.

The only issue I've noticed is the compatibility with the `usePreviewTabs` option, as mentioned above.

timkelty:
@codingbelief @benogle anything I can do to move this along?

benogle:
This needs a spec. 

cc @lee-dohm

timkelty:
@codingbelief if you merge https://github.com/codingbelief/find-and-replace/pull/2, the CI test should pass.

codingbelief:
Sorry for reply late.

codingbelief:
I'm not familiar with spec, @timkelty  can you help to deal with the new error?

timkelty:
I'm not really either :(. I can give it a shot if I get time, though.

timkelty:
This seems pretty simple, but I don't think many on this thread are familiar enough to make a spec.

Is anyone able to jump in and help? /cc @lee-dohm @benogle

lee-dohm:
I'm buried right now ... but let me see if I can find someone to chip in.

timkelty:
Thanks!

travco:
@codingbelief Pending merge in codingbelief#3 to resolve current CI failing check, and add two tests to spec for the functionality added.

codingbelief:
@travco  Thanks for you help. ～

timkelty:
@lee-dohm Can we get this merged in now that it has some tests?

lee-dohm:
Looks good to me. Thanks for contributing!

PR #549, opened 2015-10-07
Fix wrong history serialized for paths

abe33:
Not sure if it's just a copy/paste issue but I think it should serialize the `pathsHistory` and not the `replaceHistory`.

Initially reported on discuss: https://discuss.atom.io/t/remember-search-filter/21437


benogle:
oops. Thanks

PR #553, opened 2015-10-13
Increase spec resiliency by not relying on actual numbers

as-cii:
The spec changed in this PR seems to be the only failure (at least locally) for https://github.com/atom/atom/pull/8968. I couldn't backtrack the reason why this started to fail, but there seems to be a super slender change in the height of displayed items.

Manually logging out the `.selected` items revealed that, e.g. instead of `215`, item `216` gets selected: this might be related to the test runner stylesheets setup, but it seems quite negligible given that we haven't changed any line of production code in this package. As a result, I changed the spec so that it doesn't rely on _static_ numbers anymore: other than telling a better story, I think this increases its resiliency as well.

I'd :heart: to have some feedback on this before merging it.

/cc: @nathansobo @maxbrunsfeld

### DIFF
--- a/spec/results-view-spec.coffee
+++ b/spec/results-view-spec.coffee
@@ -319,15 +319,17 @@ describe 'ResultsView', ->
         itemHeight = resultsView.find('.selected').outerHeight()
         pageHeight = Math.round(resultsView.innerHeight() / itemHeight) * itemHeight
         initialScrollTop = resultsView.scrollTop()
+        itemsPerPage = Math.floor(pageHeight / itemHeight)
+        originalItemIndex = Math.floor(initialScrollTop / itemHeight) + itemsPerPage
 
         atom.commands.dispatch resultsView.element, 'core:page-up'
         expect(resultsView.find("li:last")).not.toHaveClass 'selected'
-        expect(resultsView.find("li:eq(215)")).toHaveClass 'selected'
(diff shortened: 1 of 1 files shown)


spec/results-view-spec.coffee
```diff
@@ -319,15 +319,17 @@ describe 'ResultsView', ->
         itemHeight = resultsView.find('.selected').outerHeight()
         pageHeight = Math.round(resultsView.innerHeight() / itemHeight) * itemHeight
         initialScrollTop = resultsView.scrollTop()
+        itemsPerPage = Math.floor(pageHeight / itemHeight)
+        originalItemIndex = Math.floor(initialScrollTop / itemHeight) + itemsPerPage
```
maxbrunsfeld: I had a little trouble understanding what `originalItemIndex` represents. Is it the index of the selected item before `core:page-up` is triggered? If so, maybe we should just add an assertion that the `li` at that index is initially `selected`?

maxbrunsfeld:
:+1: Great solution.

spec/results-view-spec.coffee
as-cii: Ugh, my non-native language skills strike again. The idea with this variable is to use it as a starting point for the next specs, so that we can write stuff like `originalItemIndex - itemsPerPage` after triggering a page up event.

We're already asserting that such index is selected initially a couple of lines above, so I'd say the confusion probably comes from the name of the variable. What about `selectedIndexBeforePageUp`?

PR #561, opened 2015-10-17
Add descriptions for config settings

izuzak:
As a followup to atom/atom#9096, this adds descriptions for config settings in this package.


lee-dohm:
Looks good! :rocket:

PR #572, opened 2015-10-24
Show wrap around icon when searching (close #401)

### LINKED ISSUE #401: Wrap-around notification icon
E.g. XCode and Eclipse has a nice little feature when searching a document. When the active search does a wrap-around an icon notification flashes shortly to indicate that you have reached the bottom or top and are starting over.

wrap-around

It might just be me. Otherwise I can create a PR for this small addition, probably with a more discreet notification.

mrodalgaard:
Flashes an arrow indicator for one second when find wraps around. This helps the user quickly see if he/she has reached the bottom or top. Also found in XCode, Eclipse, Mac Terminal, BBEdit, etc.

atom-wrap-around

### DIFF
--- a/styles/find-and-replace.less
+++ b/styles/find-and-replace.less
@@ -166,6 +166,25 @@ atom-workspace.find-visible {
       margin-right: @component-padding;
     }
   }
+
+  .wrap-icon {
+    @wrap-size: 96px;
+
+    display: none;
+    position: absolute;
+    left: 50%;
+    top: @wrap-size * -1.5;
+    margin-left: @wrap-size * -0.5;
+    background: rgba(136,136,136, 0.2);
+    border-radius: 10px;
+    text-align: center;
+    pointer-events: none;
+    &:before {
+      font-size: @wrap-size;
+      height: @wrap-size;
+      width: @wrap-size;
+    }
+  }
 }
 
 // Project find and replace
--- a/styles/find-and-replace.less
+++ b/styles/find-and-replace.less
@@ -168,7 +168,7 @@ atom-workspace.find-visible {
   }
 
   .wrap-icon {
-    @wrap-size: 96px;
+    @wrap-size: @font-size * 10;
 
     display: none;
     position: absolute;
@@ -176,11 +176,13 @@ atom-workspace.find-visible {
     top: @wrap-size * -1.5;
     margin-left: @wrap-size * -0.5;
     background: rgba(136,136,136, 0.2);
-    border-radius: 10px;
+    border-radius: @component-border-radius * 5;
     text-align: center;
     pointer-events: none;
     &:before {
-      font-size: @wrap-size;
+      // Octicons look best in sizes that are multiples of 16px
+      font-size: @wrap-size - mod(@wrap-size, 16px) - 16px;
+      line-height: @wrap-size;
       height: @wrap-size;
       width: @wrap-size;
     }
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -363,7 +363,7 @@ class FindView extends View
       return index if markerStartPosition.isEqual(start) and indexIncluded
       return index if markerStartPosition.isGreaterThan(start)
 
-    @showWrapIcon('icon-arrow-up')
+    @showWrapIcon('icon-move-up')
     0
 
   selectFirstMarkerBeforeCursor: =>
@@ -382,7 +382,7 @@ class FindView extends View
       markerEndPosition = marker.bufferMarker.getEndPosition()
       return index if markerEndPosition.isLessThan(start)
 
-    @showWrapIcon('icon-arrow-down')
+    @showWrapIcon('icon-move-down')
     @markers.length - 1
 
   selectAllMarkers: =>
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -24,6 +24,8 @@ class FindView extends View
       placeholderText: 'Replace in current buffer'
 
     @div tabIndex: -1, class: 'find-and-replace', =>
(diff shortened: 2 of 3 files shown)


benogle:
I like this idea and should be able to review within a week. Off the cuff: it would be nice if this kind of thing were more central so other packages could use it. Like maybe a service in the notifications package. It would also be nice if there were some text associated with it. I cant tell which arrow is supposed to be for wrapping from top, and wrapping from bottom. I'll give deeper feedback when I have time to properly review it.

mrodalgaard:
I agree this indicator type could potentially benefit other packages as a service in the [notification package](https://github.com/atom/notifications). On the other hand, the actual code used to implement this is very simple. I started out using normal notifications as a wrap around indicator, but it looked terrible.

Regarding the extra text; I do not think wrap text gives any real extra value, as you essentially know if you have clicked "Find next" you will wrap to the top and "Find previous" you will wrap to the bottom. The important part is that you know the search has wrapped around in the document.

Of course if we moved this to the [notification package](https://github.com/atom/notifications) we would have to support short text messages.

styles/find-and-replace.less
```diff
@@ -166,6 +166,25 @@ atom-workspace.find-visible {
...
+  .wrap-icon {
+    @wrap-size: 96px;
+
+    display: none;
+    position: absolute;
+    left: 50%;
+    top: @wrap-size * -1.5;
+    margin-left: @wrap-size * -0.5;
+    background: rgba(136,136,136, 0.2);
+    border-radius: 10px;
```
benogle: Can you make this a factor of the `@border-radius` var?

styles/find-and-replace.less
```diff
@@ -166,6 +166,25 @@ atom-workspace.find-visible {
       margin-right: @component-padding;
     }
   }
+
+  .wrap-icon {
+    @wrap-size: 96px;
```
benogle: maybe we should make this a factor of @font-size? @simurai what do you think?

styles/find-and-replace.less
```diff
@@ -166,6 +166,25 @@ atom-workspace.find-visible {
...
+
+    display: none;
+    position: absolute;
+    left: 50%;
+    top: @wrap-size * -1.5;
```
benogle: Any reason why this is toward the bottom of the window? I'd prefer it in the center of the window.

benogle:
I'm down to have this in just this package for now. If it proves useful, putting it in notifications and adding an API around it could make sense. 

I tested it with several themes, and it blends pretty well. I am not super fond of the arrow icons, thought. It would be nice if we had a wrap icon. I dont look at the arrow for any information other than 'it wrapped'. But I realize the octicons dont have a wrap icon.

@simurai I defer to you on more feedback and general design concerns.

simurai:
> I am not super fond of the arrow icons, though. It would be nice if we had a wrap icon.

#### Option 1

We could use the `icon-move-up` and `icon-move-down` icons. They have horizontal bars hinting that you jumped to the beginning or end.

find

#### Option 2

Or alternatively, don't show the "wrapping", but "reaching" the first/last item:

- if you do "find next", show `icon-move-down` once you reached the last item.

screen shot 2015-11-10 at 4 33 33 pm

- if you do "find previews", show `icon-move-up` once you reached the first item.

screen shot 2015-11-10 at 4 32 53 pm


#### Option 3

Was also thinking of `icon-sync` for both directions, but that's maybe a bit a stretch and people might think something "reloaded" or so.

screen shot 2015-11-10 at 4 23 38 pm

styles/find-and-replace.less
simurai: Maybe `@wrap-size: @font-size * 10;`?

When using the `icon-move-up` and `icon-move-down` (see https://github.com/atom/find-and-replace/pull/572#issuecomment-155346088), they need a bit of top/bottom padding since they take up the full height. Something like this:

```less
@wrap-size: @font-size * 10;
@wrap-icon-size: @wrap-size * 0.66;
```

and then

```less
&:before {
  font-size: @wrap-icon-size;
  line-height: @wrap-size;
  height: @wrap-size;
  width: @wrap-size;
}
```

styles/find-and-replace.less
mrodalgaard: Done https://github.com/atom/find-and-replace/commit/386cbcb7ccf93fae9d4d25513d5f368baa00a211#diff-0c91d4121c1775ed1f0c74e547086288R179

styles/find-and-replace.less
mrodalgaard: Done https://github.com/atom/find-and-replace/commit/386cbcb7ccf93fae9d4d25513d5f368baa00a211#diff-0c91d4121c1775ed1f0c74e547086288R171

I had to use modulus to keep the octicon a multiple of 16px.

styles/find-and-replace.less
mrodalgaard: I just prefer it to be closer to the find view and not directly in the users face. Since search results (at least when they are isolated) are often shown in the middle of the editor the icon view would block the user from seeing the actual result when wrapping, and thereby having to wait for the icon view to fade out.

I guess it's a matter of taste :)

mrodalgaard:
I updated the PR with better icons and relative sizing.

**Option 1**: The `icon-move-up`/`-down` icons is definitely better than just the arrow. I actually believe it improves on the normal wrap icon where you can't always see which way it wrapped.
**Option 2**: I do not think it is very intuitive and does not map well with what users would expect from other editors.
**Option 3**: Is a total no-go. This would confuse the user more than improve.

styles/find-and-replace.less
benogle: Can we move this to the center of the UI?

PR #597, opened 2015-11-17
Add an option to open results in a right or a bottom pane

### LINKED ISSUE #585: Option to display result on bottom pane
Expand the current option "Open project find results in right pane" to support opening it also in bottom pane, so:

Open project find results in additional pane = [no OR right OR bottom]

bencolon:
Convert the `openProjectFindResultsInRightPane` option to a more flexible `openProjectFindResultsInANewPane` option to open results in a new right OR bottom pane.

This update requires an Atom engine >=1.2.0 because it's using a workspace API [update](https://github.com/atom/atom/commit/1b0d72b232234cbcaf0faaf3a66c29047e594696) released from 1.2.x

About migrating the old option to the new one, I'm not sure the four lines I've added at the top of `lib/find.coffee` is the right way. It works but feel free to suggest me a better solution if needed.

Fix #585

### DIFF
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -10,6 +10,11 @@ ProjectFindView = require './project-find-view'
 ResultsModel = require './project/results-model'
 ResultsPaneView = require './project/results-pane'
 
+# To convert previous (and now unused) config setting "openProjectFindResultsInRightPane"
+if atom.config.get('find-and-replace.openProjectFindResultsInRightPane')
+  atom.config.set('find-and-replace.openProjectFindResultsInANewPane', 'right pane')
+  atom.config.unset('find-and-replace.openProjectFindResultsInRightPane')
+
 module.exports =
   activate: ({findOptions, findHistory, replaceHistory, pathsHistory}={}) ->
     atom.workspace.addOpener (filePath) ->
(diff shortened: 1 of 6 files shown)


benogle:
Cool. This needs a reviewer

acontreras89:
This has been recently proposed for the `markdown-preview` package in atom/markdown-preview#368, although for all directions.
As of now, splits for `left` and `up` cannot be achieved in these scenarios, but if we come up with something we could definitely use the same strategy in both packages and keep them consistent.

The easiest way would be to modify `Workspace`'s `open` method to be able to create new panes in those directions, but that would freeze both PRs for a bit.

bencolon:
I see your point @acontreras89 but is it really make sense to open a search results panel at the top or on the left ? Only right/bottom should cover 95% of the use cases.
And from a UX perspective, it seems weird to open a new panel "over" the code you're working on.
Thoughts ?

acontreras89:
> Only right/bottom should cover 95% of the use cases.
Agreed. I believe the PR I pointed to will be only adding options for using right/bottom panes, too.

I just wanted to cross-reference them to keep an eye on the topic, my main concern being that both PRs should share the same strategy and be consistent in the implementation. This aside, I'll be OK with pretty much any approach you end up taking.

lee-dohm:
What's the status on this?

Issue #298, opened 2014-10-10
Searching within current selection is very buggy


50Wliu:
**Scenario 1**: Let's say I want to find a phrase in a certain selection of code.  To do this, I would:
1. Select that area of code (lines 17-32)
2. Press <kbd>Ctrl</kbd>+<kbd>F</kbd>
3. Search for that phrase (`include`)
4. Press `"Only in Selection"`

*Expected result*: FnR would search in that selection and maintain the selected code
*Actual result*: It does search only in that selection but also un-highlights the selection in order to select the first match as soon as I press `"Only in Selection"`
![Atom](https://cloud.githubusercontent.com/assets/2766036/4586154/e7e5cf7a-5011-11e4-8c6b-ecb1b045d065.png)

**Scenario 2**:
1. Have FnR already set to `"Only in Selection"`
2. Select that area of code (lines 17-32)
3. Press <kbd>Ctrl</kbd>+<kbd>F</kbd>
4. Search for that phrase (`include`)

*Expected result*: FnR would display the proper number of results
*Actual result*: FnR displays `No results found`
![Atom](https://cloud.githubusercontent.com/assets/2766036/4586199/aff97020-5012-11e4-8ff6-a29640a2ab15.png)

**Scenario 3**:
1. Search for any phrase in FnR without selecting Only in Selection
2. Press Only in Selection (should display No results found)
3. Unselect Only in Selection
4. Notice how the first line that matches your phrase gets selected (actual number of results displayed)
5. Select and unselect Only in Selection
6. Notice how the second line that matches your phrase gets selected
7. Keep selecting and unselecting Only in Selection

Version 0.136.0, safe mode, Windows 7 x64.

wyand:
I think the selection should not be modified when within-selection is enabled (including when within-selection is toggled). In safe mode, this seems to be the cause of other bugs, and also has a very poor user experience:

![within-current-selection](https://cloud.githubusercontent.com/assets/1288857/6528737/1c972cce-c3f2-11e4-9918-f03e9aaa5bff.gif)

Note that when "within current selection" is toggled on, the selection in the editor is modified to select the first match, even though no actual searching operation has been performed. On the subsequent "replace all," the matches in the original selection are all modified.

As a design concern, a find and replace that has "within current selection" support should never modify the selection; it is a piece of input to the find and replace, and modifying the selection makes the find and replace workflow stateful and dependent on the order of operations.

Consider the following:

1. A user selects the top half of a large document.
[...]

PR #598, opened 2015-11-19
:bug: Fix #496 - Option buttons shouldn't move away from valid match.

### LINKED ISSUE #496: Options buttons move to next result
If I toggle the options buttons to the right in the find and replace box, the selection goes to the next occurrence. It seems to happen with all the four buttons (Regex, Case Sensitive/Insensitive, Within Current Selection, Whole Word). Atom version is 1.0.7.

colllin:
This is a proposal to fix #496.

The problem
---
Toggling the option buttons will jump you away from a selection that already matched the new settings.

The previous behavior
---
Always jump to the next match after the cursor whenever a find option is toggled.

The new behavior
---
Only jump to the next match **if none of the existing selections match** the pattern.  Inversely, **if any of the existing selections match** the pattern, it will not jump you to the next match.

Alternatives
---
I also considered and decided against these possible behaviors:

- Jump to the next match **if any of the existing selections don't match** the pattern.  Inversely, do not jump to the next match **if all of the existing selections match** the pattern.
- When a find option is toggled, destroy any selections that don't match.  If there are any remaining selections, don't jump to the next match.  Otherwise jump to the next match.

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -447,30 +447,27 @@ class FindView extends View
     else
       optionButton.removeClass 'selected'
 
-  anySelectionsMatchFindPattern: =>
-    editor = @model.getEditor()
-    selections = editor?.getSelections() or []
+  anyMarkersAreSelected: =>
+    selections = @model.getEditor()?.getSelections() or []
     _.any selections, (selection) =>
-      selectionMatchesFindPattern = false
-      editor.scanInBufferRange @model.getFindPatternRegex(), selection.getBufferRange(), ({range}) ->
-        selectionMatchesFindPattern = selectionMatchesFindPattern or _.isEqual(range, selection.getBufferRange())
-      selectionMatchesFindPattern
+      _.any @model.markers or [], (marker) =>
+        _.isEqual(marker.getBufferRange(), selection.getBufferRange())
 
   toggleRegexOption: =>
     @search(useRegex: not @model.getFindOptions().useRegex)
-    @selectFirstMarkerAfterCursor() unless @anySelectionsMatchFindPattern()
+    @selectFirstMarkerAfterCursor() unless @anyMarkersAreSelected()
 
   toggleCaseOption: =>
     @search(caseSensitive: not @model.getFindOptions().caseSensitive)
-    @selectFirstMarkerAfterCursor() unless @anySelectionsMatchFindPattern()
+    @selectFirstMarkerAfterCursor() unless @anyMarkersAreSelected()
 
   toggleSelectionOption: =>
     @search(inCurrentSelection: not @model.getFindOptions().inCurrentSelection)
-    @selectFirstMarkerAfterCursor() unless @anySelectionsMatchFindPattern()
+    @selectFirstMarkerAfterCursor() unless @anyMarkersAreSelected()
 
   toggleWholeWordOption: =>
     @search(@findEditor.getText(), wholeWord: not @model.getFindOptions().wholeWord)
-    @selectFirstMarkerAfterCursor() unless @anySelectionsMatchFindPattern()
+    @selectFirstMarkerAfterCursor() unless @anyMarkersAreSelected()
 
   updateReplaceEnablement: ->
     canReplace = @markers?.length > 0
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -449,11 +449,11 @@ class FindView extends View
 
   anySelectionsMatchFindPattern: =>
     editor = @model.getEditor()
-    selections = editor?.getSelections() || []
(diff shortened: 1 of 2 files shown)


colllin:
Screen captures...

Before: always jumps even if existing selection already matched.
before

After: doesn't jump if existing selection matches.
after-nojump

After: jumps if existing selection doesn't match.
after-jump

50Wliu:
I like this.  Refs #298.

lib/find-view.coffee
```diff
@@ -447,21 +447,30 @@ class FindView extends View
     else
       optionButton.removeClass 'selected'
 
+  anySelectionsMatchFindPattern: =>
+    editor = @model.getEditor()
+    selections = editor?.getSelections() or []
+    _.any selections, (selection) =>
+      selectionMatchesFindPattern = false
+      editor.scanInBufferRange @model.getFindPatternRegex(), selection.getBufferRange(), ({range}) ->
```
benogle: This scan seems unnecessary. The model should know if there is a current result marker IIRC. Seems like we should just be able make the decision based on the existence of the current result marker.

benogle:
I'm down with this functionality. Need someone to look over all the code still though.

lib/find-view.coffee
```diff
@@ -447,21 +447,27 @@ class FindView extends View
     else
       optionButton.removeClass 'selected'
 
+  anyMarkersAreSelected: =>
+    selections = @model.getEditor()?.getSelections() or []
+    _.any selections, (selection) =>
+      _.any @model.markers or [], (marker) =>
+        _.isEqual(marker.getBufferRange(), selection.getBufferRange())
```
benogle: I'm not keen on the underscore usage here. If there are a ton of markers this will probably break down. @maxbrunsfeld  @nathansobo is there a nice way to check if any markers are selected?

maxbrunsfeld: @colllin I think that we need to avoid a `selections.length x markers.length` loop here, mainly because there can be hundreds of thousands of markers. I think the performance will be more predictable if you query for the markers by range, using `TextEditor::findMarkers`. There is already code to do this in [`BufferSearch.prototype.findMarker`](https://github.com/atom/find-and-replace/blob/d719c79b358cf59e15ca185091965d29a9ead49b/lib/buffer-search.coffee#L194). You could use that method like this:

```coffee
for selectedRange in @model.getEditor().getSelectedBufferRanges()
  if @model.findMarker(selectedRange)
    return true
```

spec/find-view-spec.coffee
```diff
@@ -844,6 +844,37 @@ describe 'FindView', ->
           atom.commands.dispatch(findView.findEditor.element, 'core:confirm')
           expect(findView.descriptionLabel).toHaveClass 'text-error'
 
+      describe "when there are existing selections", ->
+        describe "when any existing selections match the pattern", ->
+          it "shouldn't jump to the next match when toggled on", ->
```
maxbrunsfeld: Usually we avoid the `should` phrasing in the specs, opting for the slightly more direct `it "does not jump to the next match when toggled on"`.

spec/find-view-spec.coffee
```diff
@@ -892,6 +956,39 @@ describe 'FindView', ->
...
+
+          it "should jump to the next match when toggled off", ->
+            # It's impossible to create a test where `caseSensitive: true` matches but `caseSensitive: false` doesn't.
+            # Instead we'll create a generic selection and ensure that it jumps to the next match.
+            findView.model.setFindOptions caseSensitive: true
+            findView.findEditor.setText 'WORDs'
+            editor.setSelectedBufferRange [[0, 0], [0, 5]]
+            findView.caseOptionButton.click()
+            expect(editor.getSelectedBufferRange()).toEqual [[1, 0], [1, 5]]
+
```
maxbrunsfeld: I usually don't say this, but I think there might be a few *too* many tests for this code change. They look very good, but I would probably combine the tests for `when toggled on` and `when toggled off` into one, in which you click the button twice, and check that the selection persists at every stage.

maxbrunsfeld:
Thanks for figuring this stuff out @colllin. I left a few suggested changes, but overall this looks great.

colllin:
Thank you for the suggestions and the tip about `Array.prototype.some`.

I...
- changed the `anyMarkersAreSelected` logic to use `BufferSearch.prototype.findMarker`
- combined the toggled on/off tests
- changed the spec phrasing to avoid "should"

maxbrunsfeld:
This is a great PR, @colllin. Thanks so much for sticking with it.

maxbrunsfeld:
This will be in Atom 1.4.0 and 1.4.0-beta0.

Issue #165, opened 2014-03-20
Find and replace with preserve case


jasonkim:
When you search for abc (case insensitive) and want to replace it with xyz, it should match on Abc and replace it with Xyz. This is a feature that I found useful in sublime.

claaslange:
:+1: for the simple case as well

orospakr:
Emacs also has this behaviour, by default, and it's profoundly useful.

littlebee:
+1 very useful  intelliJ feature.  I miss.    Here is the use case, it is not, imnsho, an "edge" case considering we frequently, for multiple languages, mix the case to indicate something is a class or a var.  I want to replace 'collection' below with 'model' without doing 2 f&r:

```coffeescript
  ###
    Collection component provides a collection context to all children
  ###
  class x.Collection extends React.Component
    displayName: "widgets.react.Collection"

    @propTypes:
      collection: React.PropTypes.oneOfType([
        React.PropTypes.instanceOf(Backbone.Collection)
        React.PropTypes.func
      ]).isRequired
      fetch: React.PropTypes.bool
      fetchOptions: React.PropTypes.object
```

Rosseyn:
Hands down most awaited feature.

BigWillie:
+ 1 - this would be killer to have!

keomaborges:
+1 - I still have to open Sublime to do this kind of replace.

adambuczynski:
This issue was opened almost 2 years ago, it would be great to have it implemented.

I too would find it very useful, since in my Angular projects I often need to duplicate existing modules and then rename their name to something else. This affects not only the module name, but also the controller name, template name, etc. And these could all be in different cases, for example camelCase, PascalCase, or just simply lowercase. Currently, this would require me three find & replace operations.

If there'd be a toggle to trigger case-aware replacement, that would save a lot of time.

Bonus points if support for snake_case or kebab-case (dasherized) strings can also be added. This would probably involve detection of what case the search term and replacement term are in (and I think both should match). Then, if doing a find & replace with the toggle turned on, all other case variants could be generated from the search term and found & replaced by the equivalently cased replacement term.

ronnyroeller:
+1 for the simple case

siliconjohn:
+1 - I still have to open Sublime to do this kind of replace.

mliq:
+1

diddierh:
+1 I still have to use subl for this simple task

hillam:
+1

[...]

PR #606, opened 2015-11-28
Disconnect SelectNext from findOptions and vary behavior based on initial selection.

### LINKED ISSUE #521: Please separate "whole word" option in CMD+D vs Find cases
These aren't the same thing to me and it seems very odd that they're the same option.  Personally I like how CMD+D works with this option on (it searches for a whold word if and only if you have a whole word selected).  But I almost never use this option when opening a find dialog.

Personally, I think the CMD+D case is personal preference -- Either you expect it to select a whole word or not and you expect it to act like that globally.  I'd put that option in the package settings.

colllin:
This PR fixes #521 and reverts #482 and #541.

`find-and-replace:select-next` has two core functions.  When nothing is selected, it performs the **SelectWord** function and selects the word(s) under the cursor(s).  When a selection exists, it performs the **SelectNext** function and selects the next occurrence of the selected text.

As discussed in #521, this PR changes the behavior of **SelectNext** depending on whether the existing selection was made using **SelectWord**.

- If the existing selection was made using **SelectWord**, then following **SelectNext** actions will match only whole words.
- If the existing selection was made any other way (e.g. mouse, shift + arrow keys, find panel), then following **SelectNext** actions will match any occurrence.

Since `find-and-replace:select-all` piggybacks on `select-next`'s logic, this PR also affects the behavior of `select-all`.  Essentially, `select-all` is the equivalent of dispatching `select-next` until all occurrences are selected.

Demo:
select-next

I believe this functionality was also suggested in #430.

### DIFF
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -125,7 +125,7 @@ module.exports =
       editor = editorElement.getModel()
       selectNext = @selectNextObjects.get(editor)
       unless selectNext?
-        selectNext = new SelectNext(editor, {@findOptions})
+        selectNext = new SelectNext(editor)
         @selectNextObjects.set(editor, selectNext)
       selectNext
 
--- a/lib/select-next.coffee
+++ b/lib/select-next.coffee
@@ -1,5 +1,5 @@
 _ = require 'underscore-plus'
-{Range} = require 'atom'
+{CompositeDisposable, Range} = require 'atom'
 
 # Find and select the next occurrence of the currently selected text.
 #
@@ -8,7 +8,7 @@ module.exports =
 class SelectNext
   selectionRanges: null
 
-  constructor: (@editor, {@findOptions}) ->
+  constructor: (@editor) ->
     @selectionRanges = []
 
   findAndSelectNext: ->
@@ -51,7 +51,14 @@ class SelectNext
 
   selectWord: ->
     @editor.selectWordsContainingCursors()
-    @wordSelected = @isWordSelected(@editor.getLastSelection())
+    lastSelection = @editor.getLastSelection()
+    if @wordSelected = @isWordSelected(lastSelection)
+      disposables = new CompositeDisposable
+      clearWordSelected = () =>
+        @wordSelected = null
+        disposables.dispose()
+      disposables.add lastSelection.onDidChangeRange clearWordSelected
+      disposables.add lastSelection.onDidDestroy clearWordSelected
 
   selectAllOccurrences: ->
     range = [[0, 0], @editor.getEofBufferPosition()]
@@ -74,20 +81,16 @@ class SelectNext
   addSelection: (range) ->
     selection = @editor.addSelectionForBufferRange(range)
     @updateSavedSelections selection
-    disposable = selection.onDidDestroy =>
-      @wordSelected = null
-      disposable.dispose()
 
   scanForNextOccurrence: (range, callback) ->
     selection = @editor.getLastSelection()
     text = _.escapeRegExp(selection.getText())
 
-    @wordSelected ?= @isWordSelected(selection)
     if @wordSelected
       nonWordCharacters = atom.config.get('editor.nonWordCharacters')
       text = "(^|[ \t#{_.escapeRegExp(nonWordCharacters)}]+)#{text}(?=$|[\\s#{_.escapeRegExp(nonWordCharacters)}]+)"
 
-    @editor.scanInBufferRange new RegExp(text, (if @findOptions.caseSensitive then 'g' else 'gi')), range, (result) ->
+    @editor.scanInBufferRange new RegExp(text, 'g'), range, (result) ->
       if prefix = result.match[1]
         result.range = result.range.translate([0, prefix.length], [0, 0])
       callback(result)
@@ -117,7 +120,7 @@ class SelectNext
     @isNonWordCharacter(@editor.getTextInBufferRange(range))
 
   isWordSelected: (selection) ->
-    if @findOptions.wholeWord and selection.getBufferRange().isSingleLine()
+    if selection.getBufferRange().isSingleLine()
       selectionRange = selection.getBufferRange()
       lineRange = @editor.bufferRangeForBufferRow(selectionRange.start.row)
       nonWordCharacterToTheLeft = _.isEqual(selectionRange.start, lineRange.start) or
--- a/lib/select-next.coffee
+++ b/lib/select-next.coffee
@@ -54,7 +54,7 @@ class SelectNext
     lastSelection = @editor.getLastSelection()
     if @wordSelected = @isWordSelected(lastSelection)
       disposables = new CompositeDisposable
-      clearWordSelected = () =>
+      clearWordSelected = =>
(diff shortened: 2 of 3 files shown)


benogle:
Thanks, hopefully this will be put to rest once and for all. 

Also, I (we!) appreciate your initiative, attention to detail and patience with find and replace things. :heart:

colllin:
Thank you! :) !

I hope it's a workable solution for people.  It's interesting how
challenging these simple-seeming design problems can be.  I think some of
the confusion comes from the behavior of the FNR markers, and I'm playing
with it locally to see if I can find a good solution.  I'll let you know if
I do.

jeancroy:
With this merged, we should reopen https://github.com/atom/atom/issues/9003
Because sublime text definitively get caseSensitive option from findOptions.
One can refers to screenshot [here](https://github.com/atom/find-and-replace/issues/521#issuecomment-153903555) for behavior

Not a bad fix, just disconected too many things.

colllin:
I agree with the findOptions being connected in some situations but not all situations. 

I've spent some time thinking about it, and I think that the behavior should diverge depending on whether the find panel was used to make the selection. If the find panel initializes the selection, for example by pressing `cmd-g`, then presses of `cmd-d` should select the next marker from the find panel, which uses the find options. If the section is made any other way, then I think `cmd-d` should ignore the find options and use the current behavior.

Another thing to address is that you can't press `cmd-d` when the find panel is focused.

jeancroy:
As thing are right now, initial selection can totally describe whole-word option.
However that remove the option to choose case sensitivity.


It can get more fancy I guess. But I wanted to point this was a difference from doing like sublime.

yamsellem:
@colllin @jeancroy Sorry to bother, but:
— in which release of atom this has been merged?
— how do we selected something "using SelectWord" to get the whole word match?

I'm using atom 1.5.4 and still having problem with this.
Thanks.

colllin:
@yamsellem 

If you make the initial selection using `cmd-d` to select the whole word surrounding your cursor, then further presses of `cmd-d` will add the next *whole word* matches to your selection:

Screenshot when initial selection used `cmd-d`, and then pressed `cmd-d` two more times:
image

If you make the initial selection any other way (double-click, click-drag, shift-right-right-right, shift-option-left), or if the existing selection is not a whole word, then pressing `cmd-d` will add any string match to your selection.

Screenshot when initial selection used shift-arrows, and then pressed `cmd-d` three more times:
image

Does that help?  What does no mouse/keyboard/find mean?

yamsellem:
@colllin That's perfect! I wasn't aware I could directly select the closest to cursor word with `cmd-d`. 

I dislike the other way around (not the whole word).
Is there a way to make every action with the mouse or keyboard (_except the find and replace panel_) to do a whole word select?

yamsellem:
@collin Sorry to bother, but something seems weird to me.
`cmd-d` is also capturing strings when we are selecting variables.

cursor

How can we refactor a variable name with this behavior? 
I know this is consistent with sublime text, but it still seems weird to me.

Is there a way to only select a given variable with that shortcut?
Thanks again.

colllin:
Not that I know of, but it seems like a reasonable feature. I've seen that feature in IntelliJ idea , I think. A lighter solution might be to give users a way to skip the word that was just selected. For example, based on your screenshot, after pressing `cmd-d` twice, the first two instances of `value` would be selected in line 2. Then, pressing a different shortcut, maybe `alt-cmd-d`, would select the third instance and delete the second selection. So you would be left with two selections again, now just the variable names.

I would definitely use a feature like that. Either way, this is technically a new feature request. You could submit a feature request as a new issue to ask if the maintainers would be open to a pull request, or if they'd recommend implementing it as a package instead. If they're open to a pull request, let me know if you need help.

yamsellem:
@collin on a `cmd-d` selection, `cmd-k` skip the actual occurrence ;-)
So this help, but 'select every occurrence' of a variable would be very useful. Like 'extract method' or 'inline variable'. Refactoring needs to find its way in atom ^^.

jeancroy:
> Refactoring needs to find its way in atom 
I'd keep an eye on https://github.com/tree-sitter/tree-sitter

From what I understand atom plan to transition to a real language parsing back end instead of current regex based solution. When / If that happens, things like refactoring will be much easier to implement in a robust manner. Also auto indent and various code transform.

bostrom:
Can I ask what the motivation is behind the decision to make cmd-d (when initiated by mouse/arrow keys, etc) perform a case sensitive select-next? 

Coming from Sublime, I often use the arrow keys to select part of (or a whole) a word/variable and selecting all relevant occurrences with cmd-d (select-next) and cmd-k + cmd-d (skip current, select next). This works well since Sublime uses case insensitive find and would include all possible matches, and I would just cmd-k past the ones that I don't want to include. Cmd-u is also a good companion here, undoing the selections in reverse order.

Moving to Atom, I find that it has all of those nice keyboard shortcuts, but that it breaks the inclusive behavior by performing case insensitive select-next with cmd-d. This leaves out possibly relevant matches that would otherwise be easily skippable with cmd-k. 

I agree that the case where the select-next is initiated from mouse/arrow keys may well be decoupled from the find panel's options since they're not even visible while doing cmd-d. Also, the case when initiating the select-next by using cmd-d to select the word under the cursor could be case insensitive. 

But I'd rather see cmd-d using mouse/arrow keys be more inclusive by making it case insensitive. 

Thoughts?

colllin:
@bostrom – I didn't know about the cmd-k and cmd-u shortcuts.  Thanks for that tip.  I'll try it and see how that affects how my understanding of cmd-d / SelectNext.  

I'm having trouble understanding a use case for selecting multiple case-insensitive matches, though.  Can you suggest one to help me better understand your usage?

bostrom:
@colllin A common use case of mine is fixing typos or renaming (part of) variables written in different styles across a file. 

Doing this in Sublime also requires the help of the "Replace preserve case" functionality of Sublime's [MultiEditUtils plugin](https://github.com/philippotto/Sublime-MultiEditUtils/tree/1.6.4#preserve-case-while-editing-selection-contents). The docs provide some examples and explain the rationale:

> When multi-selecting all occurences of an identifier it is cumbersome to change it to another one if the case differs (camelCase, PascalCase, UPPER CASE and even cases with separators like snake_case, dash-case, dot.case etc.). The "Preserve case" feature facilitates this. Just invoke "Preserve case" via the command palette (or define an own keybinding) and type in the new identifier.

This functionality is not yet available in Atom (AFAIK), but there seems to be a request open for that as well (https://github.com/atom/find-and-replace/issues/165). This replace functionality could arguably be part of a different plugin, but a case insensitive multi-select would still be required to make this work smoothly. 

A case insensitive matching would be more versatile as it still would be easy to skip irrelevant matches with `cmd-k + cmd-d` or undoing with `cmd-u`.

crccheck:
did this get reverted? "Select-Next" is always acting like whole words is off now, Atom 1.14.1. There's no docs in the files changed of this PR and nothing on https://atom.io/docs/api/search/latest?utf8=%E2%9C%93&q=select-next

PR #607, opened 2015-11-28
Fix #591 - ReplaceNext shouldn't skip current selection.

### LINKED ISSUE #591: Find and Replace starts with the second result of the search, instead of the first
Example:

> Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s

1. Double-click the first word "Lorem" to select it.
2. CMD+F
3. Enter some other text in the "Replace" field
4. Press enter

Result - replaces the SECOND occurrence of the word "Lorem", instead of the one that you selected in step 1.
Expected result - must replace the first word that you originally selected in step 1, and then proceed to highlight the second one.

I don't know if this is intended behaviour, but it felt unnatural to me having to constantly go back to the first result of the search and manually replace it.

Cheers

colllin:
Before:
before

After:
after

Fixes #591


spec/fixtures/sample.js
```diff
@@ -10,4 +10,4 @@ var quicksort = function () {
   };
 
   return sort(Array.apply(this, arguments));
-};
\ No newline at end of file
+};
```
benogle: Can you remove this change?

benogle:
Other than the sample change, this LGTM.

benogle:
Thanks for fixing this btw :heart:

colllin:
Thanks for the review!  I also noticed a function called `selectFirstMarkerStartingFromCursor` the other day, so I renamed the `firstMarkerIndexFromCursor` function to `firstMarkerIndexStartingFromCursor` to be consistent with that name.

PR #611, opened 2015-12-03
Fix up wrap around icon

benogle:
This is an extension on https://github.com/atom/find-and-replace/pull/572. There were a couple issues:

* The icon was only shown when the find view was open
* The icon could not be centered vertically over the editor in CSS because it was a child of the find view
* The icon was centered horizontally over the fnr panel, not the editor, so it didnt make sense with splits
* The colors didnt work well with some theme combinations

Looks like this now:

screen shot 2015-12-03 at 3 29 13 pm

cc @mrodalgaard


benogle:
Travis is soooo sloooow.

PR #613, opened 2015-12-07
Only show search wrap icon when scroll position changed

maxbrunsfeld:
I think @nathansobo, @kevinsawicki and I have all found it annoying that the icon shows up even if the editor does not scroll.


nathansobo:
Awesome. Thanks for jumping on this!

maxbrunsfeld:
The build is failing due to a bug in v1.2.3 that is fixed on `beta`. I think I'll just wait to ship this until we roll the railcars on Wednesday.

PR #614, opened 2015-12-07
Add service to access result marker layer

nathansobo:
This PR adds a new `find-and-replace@0.0.1` service that exposes a single method `resultsMarkerLayerForTextEditor`. Give it a `TextEditor`, and it will return a marker layer used to store the search results for that editor. It returns null in versions of Atom prior to 1.3.0.

This should make it straightforward to fix packages that relied on finding find-and-replace markers in the default marker layer by simply wrapping introducing a service call to retrieve the appropriate layer for that editor before calling `findMarkers`. Everything else should remain the same.

I'm planning to hotfix this into beta tomorrow morning so it goes out with the introduction of layers.

Refs atom/atom#9208
/cc @abe33 @lee-dohm

### DIFF
--- a/lib/buffer-search.coffee
+++ b/lib/buffer-search.coffee
@@ -58,7 +58,7 @@ class BufferSearch
 
   resultsMarkerLayerForTextEditor: (editor) ->
     unless layer = ResultsMarkerLayersByEditor.get(editor)
-      layer = editor.addMarkerLayer()
+      layer = editor.addMarkerLayer?()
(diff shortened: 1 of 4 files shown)


lee-dohm:
Looks good to me!

abe33:
:+1:

It'll be so much simpler to not having to care about the active editor!

:clap:

nathansobo:
Keep in mind that the service will return `null` instead of a layer for versions before 1.3.0.

PR #616, opened 2015-12-08
automatic search within selection

ivoadf:
The objective of this enhancement is to automatically activate search within selection when find-and-replace is called and the selected text has one or more '/n' if no newline is found the selected text is used as the search pattern as it was implemented before.

Would love to have feedback both on my implementation (which is buggy, but at the moment I don't know how to fix it) and on the idea.

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -479,12 +479,12 @@ class FindView extends View
     @selectFirstMarkerAfterCursor() unless @anyMarkersAreSelected()
 
   activateSelectionOption: =>
-    @search(@findEditor.getText(),inCurrentSelection: true)
+    @search(@findEditor.getText(), inCurrentSelection: true)
     @model.getFindOptions().inCurrentSelection=true
     @selectFirstMarkerAfterCursor() unless @anyMarkersAreSelected()
 
   deactivateSelectionOption: =>
-    @search(@findEditor.getText(),inCurrentSelection: false)
+    @search(@findEditor.getText(), inCurrentSelection: false)
     @model.getFindOptions().inCurrentSelection=false
     @selectFirstMarkerAfterCursor() unless @anyMarkersAreSelected()
 
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -479,15 +479,13 @@ class FindView extends View
     @selectFirstMarkerAfterCursor() unless @anyMarkersAreSelected()
 
   activateSelectionOption: =>
-    @search(@findEditor.getText(), inCurrentSelection: true)
-    @model.getFindOptions().inCurrentSelection=true
-    @selectFirstMarkerAfterCursor() unless @anyMarkersAreSelected()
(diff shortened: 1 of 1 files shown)


ivoadf:
I believe I have fixed all the issues and the changes are working as intended. Would like some feedback on the idea.

lee-dohm:
I have a couple concerns:

* I don't feel like the functionality is obvious enough that it is choosing to search within the selection on the user's behalf. 
* How is one supposed to discover that stretching a selection past a newline is the magic that makes this feature work?

ivoadf:
Here is my point of view:
The idea behind activating search within selection when multiple lines are selected was related to the fact that when only one line is selected it is used as the search pattern but multiple lines selected have no effect. In my opinion it is intuitive that when you select text and activate find-and-replace it should have some effect like using that text as the search pattern or searching only within the selected text.
And in this case the condition is if the text selected has multiple lines or not.

lee-dohm:
[An "intuitive" interface is one that is familiar.](http://ux.stackexchange.com/a/1284) As StackExchange answer points out, there is what someone knows and what they need to know. If someone is already familiar with a find and replace system that works the way you're suggesting ... great. But I've never used one that works this way, so I'm going to guess that I'm not alone. What I'm asking is ... if someone isn't already familiar with this metaphor, how are they going to pick it up without being frustrated?

ivoadf:
I understand what you are saying, but I don't have an answer that I think fits the purpose.
My first thought was to add something like, *Select multiple lines* to the label that appears when you hover the **only in selection** button but that can be confusing as well.

Would be happy to ear out some suggestions and to try and implement them.

yufanyufan:
I personally find the proposed behavior very natural, and it is how most editor works. For example, this is exact how Visual Studio code works.

lib/find-view.coffee
```diff
@@ -208,8 +208,12 @@ class FindView extends View
   focusFindEditor: =>
     selectedText = atom.workspace.getActiveTextEditor()?.getSelectedText?()
     if selectedText and selectedText.indexOf('\n') < 0
+      @deactivateSelectionOption()
       selectedText = Util.escapeRegex(selectedText) if @model.getFindOptions().useRegex
       @findEditor.setText(selectedText)
+    else if selectedText and selectedText.indexOf('\n') > 0
+      @activateSelectionOption()
+      @findEditor.setText("")
```
yufanyufan: Here, it should not clear the Text.
This is a natural workflow for search in selection.
1. select the word to search, then ctrl-f.
2. select the paragraph, then ctrl-f again. Now search in selection is activated, and text to search is not changed.

lee-dohm:
Thanks for the interest and the work you put in on implementing this.

We discussed this as a team and we agree that [my original concerns](https://github.com/atom/find-and-replace/pull/616#issuecomment-164096138) about this feature being hard to discover and confusing still apply. Because of this, we won't be adding this feature.

Thanks for your interest in Atom.

PR #628, opened 2016-01-01
Add package option to disable search-wrap icon

Alhadis:
The wrap-around icon added in #572 is a nice touch, but it might be (slightly) distracting to some users... I've added an option to control the visibility of the visual cue. It's enabled by default; turning it off will hide the overlay.

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -522,6 +522,7 @@ class FindView extends View
     @wrapIcon = $(wrapIcon)
 
   showWrapIcon: (icon) ->
+    return unless atom.config.get('find-and-replace.showSearchWrapIcon')
     editor = @model.getEditor()
     return unless editor?
     editorView = atom.views.getView(editor)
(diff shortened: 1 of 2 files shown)


benogle:
Cool. I know @maxbrunsfeld wanted this as well

maxbrunsfeld:
Yeah, the icon is currently very obtrusive, as it appears whenever the search wraps around, even if the scroll position did not change. As of https://github.com/atom/find-and-replace/pull/613, it only shows up if the scroll position changes. This prevents the annoying case where there are only a few search results in one spot, and the icon shows up repeatedly. I'm personally happy w/ this behavior, but I guess I am fine with including this setting in addition.

@Alhadis do you feel that the option is still necessary if the icon only shows up when the editor scrolls?

Alhadis:
@maxbrunsfeld Aye, I do. To be honest, I don't care whereabouts something matched with relation to the window's starting point.

Besides, I can always use the editor's scroll-bar to determine if the position's changed.

benogle:
Thanks!

Alhadis:
No, thank YOU. :D

PR #640, opened 2016-01-11
Project find panel default

csholmq:
Changed default setting to use project find *side* panel.

Standard behaviour is having a tab created for every project find performed. It then gets focus taken off of it everytime you click one of your results. I propose to use side panel as default. One that stays visible when clicking one of the matches.


benogle:
This used to be the default, and people were not into it, so it defaults to the active pane. We're inclined to keep it as it is currently. Sorry about that.

csholmq:
I see. I guess I can accept the vote of the majority, even though I don't understand how one can prefer that way of working :)

PR #643, opened 2016-01-13
Add menu command to clear find/replace history.

boone:
Sometimes I search and replace sensitive info and I don't want to leave it laying around on the system. I added a menu item to allow a user to clear the find/replace history data.


50Wliu:
Makes sense.

benogle:
Thanks!

PR #644, opened 2016-01-14
Open file in pending state on single click

kuychaco:
Single click on search result opens file in pending state. Double click opens file in regular non-pending state.

Corresponds to [atom/atom/pull/10178](https://github.com/atom/atom/pull/10178)
Addresses [atom/atom/issues/9165](https://github.com/atom/atom/issues/9165)

Specs included.

### DIFF
--- a/lib/project/match-view.coffee
+++ b/lib/project/match-view.coffee
@@ -42,12 +42,13 @@ class MatchView extends View
       @replacementText.text('').hide()
       @matchText.removeClass('highlight-error').addClass('highlight-info')
 
-  confirm: ->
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
:ship:

adambuczynski:
@kuychaco this seems broken in Atom 1.10.0-dev-748dbcd. Does this test still pass? 

https://github.com/atom/find-and-replace/blob/66f09c532bb4f7b941282b99d4daf85a08d2288c/spec/results-view-spec.coffee#L436

See also https://github.com/atom/find-and-replace/issues/680

PR #645, opened 2016-01-14
Focus project-find search input upon toggle

felixjung:
This PR adds the following change: when toggling the project-find panel via the ``project-find:toggle`` command from an invisible state, the search input is focused after the panel becomes visible. This is preferable for the following reasons:

* This exact behaviour is the default for the ``find-and-replace:toggle`` command. Enabling the same behaviour for ``project-find:toggle`` makes find-and-replace more consistent from a UX perspective.
* Some people prefer to use toggling instead of showing and dismissing via ``Esc`` and create according key bindings in their ``keymap.cson``. In such cases it's convenient to also focus the search input.

### DIFF
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -60,19 +60,11 @@ module.exports =
 
     @subscriptions.add atom.commands.add 'atom-workspace', 'project-find:show', =>
       @createViews()
-      @findPanel.hide()
-      @projectFindPanel.show()
-      @projectFindView.focusFindElement()
+      showPanel(@projectFindPanel, @findPanel, @projectFindView.focusFindElement)
 
     @subscriptions.add atom.commands.add 'atom-workspace', 'project-find:toggle', =>
       @createViews()
-      @findPanel.hide()
-
-      if @projectFindPanel.isVisible()
-        @projectFindPanel.hide()
-      else
-        @projectFindPanel.show()
-        @projectFindView.focusFindElement()
+      togglePanel @projectFindPanel, @findPanel, @projectFindView.focusFindElement
 
     @subscriptions.add atom.commands.add 'atom-workspace', 'project-find:show-in-current-directory', ({target}) =>
       @createViews()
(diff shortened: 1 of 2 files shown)


lib/find.coffee
```diff
@@ -72,6 +72,7 @@ module.exports =
         @projectFindPanel.hide()
       else
         @projectFindPanel.show()
+        @projectFindView.focusFindElement()
```
lee-dohm: Rather than just adding the one missing line, maybe we could turn these two implementations into a common function so that find and find-in-project always work the same?

felixjung: Sure, that would make sense. I guess in that case we can do the same for other functions like ``show``. Most of these functions seem to share some actions:

* Create the view.
* Disable 'the other find panel'.
* Show/hide the correct find panel.
* Optionally focus a specific element.

I would propose the following changes:

1. Write a function that determines 'the other find panel' based on the panel to be activated.
2. Generalize functions so that they take a panel to activate and an optional action like 'focus input x' as arguments.

What do you think? I'll try to write some more code soonish.

felixjung:
I've generalized the ``toggle`` and ``show`` functions for the two panels. I think I stumbled upon a scoping bug for the project-find view. Functions were defined with ``->`` rather than ``=>`` and thus the focusing did not work (see d550e62 for details).

lib/find.coffee
```diff
@@ -60,18 +60,11 @@ module.exports =
 
     @subscriptions.add atom.commands.add 'atom-workspace', 'project-find:show', =>
       @createViews()
-      @findPanel.hide()
-      @projectFindPanel.show()
-      @projectFindView.focusFindElement()
+      showPanel @projectFindPanel, @findPanel, @projectFindView.focusFindElement
```
50Wliu: :shirt: Some extra whitespace at the end of this line

felixjung:
How sloppy...

felixjung:
Do these functions need unit tests?

lee-dohm:
It's preferable if you can add unit tests for these, yes. That way we can be relatively certain we don't accidentally break them as things move forward.

felixjung:
I will try to figure out how to write these UI tests based on the existing tests for the package. Thanks @lee-dohm.

benogle:
Thanks!

felixjung:
But the tests were still missing o_O

lee-dohm:
It would be great if you could submit a separate PR for the tests :grinning:

felixjung:
Ahahahah, will try. Thanks.

PR #663, opened 2016-02-06
Reuse existing marker layers

kuychaco:


kuychaco:
Decided not to serialize marker layers for find-and-replace (since `maintainHistory` is false) and therefore we will create new marker layers each time. See https://github.com/atom/text-buffer/pull/130

PR #673, opened 2016-03-01
Add support for new pane pending item api

kuychaco:
Corresponding pane pending item API - https://github.com/atom/atom/pull/10959

Signed-off-by: Michelle Tilley <binarymuse@github.com>


kuychaco:
Fixes atom/atom#11007

Issue #347, opened 2015-02-11
UX: no close button


mark-hahn:
It is a pain to have to go to the keyboard to close the find/replace dialog and then back to the mouse, especially for those of us who use the mouse on the left. Many other atom dialogs also have this non-standard UX. I can't remember any dialogs in any other app that don't have close buttons.

jerone:
:+1: As a mostly mouse/UI/non-keyboard user, I'm missing this simple UX too.

itsderick:
I agree with this as well. 👍

I should have known but I couldn't even figure out how to close it with a keyboard until after a few seconds of fiddling around with different key bindings. Finally, a sentence showed up that said I could close it with Esc after hitting Command + Shift + F. A close button might be more helpful, or also add that sentence when it's opened with just Command + F.

Great job guys on this editor guys. 😄

BramVanroy:
I don't know if it's a habit to +1 posts that you agree with on GitHub, but I definitely +1 this. I'd like to see it implemented as it is with the JSLint package (top right):

![JSLint closing button](https://cloud.githubusercontent.com/assets/2779410/8396451/2943495e-1da6-11e5-9c78-db18c83a3ca8.PNG)

From left to right: sort lines (n/a for search), minimise pane, close pane.

mark-hahn:
> but I definitely +1 this

A +1 by itself is considered noise and not OK.  However your post is not
noise but contains information.  A slide was shown at codeconf with pages
and pages of just +1.

The Atom forum at discuss.atom.io blocks you from posting less than 10
chars (used to be 20 but I bitched).

BramVanroy:
So because my post contained information (the image) it's OK, but if I had simply posted *+1* it wouldn't? Thanks, now I know that for the future.

joshuapack:
In the latest 1.7.4, I have been unable to close the find panel with esc. Not sure if plugin issue or something else. A graphical X would help in cases like this. I have to keep it there until I have time to figure out if it is a plugin or not.

SystemLogoff:
Plus one on this.

reggieb:
+1

tuffylock:
Users should be able to use keyboard-only AND mouse-only navigation *in as many situations as possible*. 

Switching from mouse to keyboard is *just as annoying as switching from keyboard to mouse*. 

The top of the find dialog is **giant** and **practically empty**. 

There's no excuse for not having a close button.

ovokuro:
+1

[...]

Issue #164, opened 2014-03-20
No way to hide the Find/Replace Pane


jamiepeloquin:
I can not seem to find a way to get rid of the Find/Replace pane in the editor. I have tried various shortcuts (⌘P, ⌘T , ⌘⇧F) and options via Find menu. So far, the only way to close the Find/Replace pane is to close the entire window.

reggieb:
Just started using Atom. Not having a visible way to close the search panel was a pain. Googling got be the Esc option, but I know I'm going to have to do that google again some time (I'm getting old and don't have a great memory). I hate systems where you need to know 'magic' keystrokes to do anything. A small x as suggested here takes up minimal screen space, and in common practice in more windowed user interfaces. 

As for users learning the key command - they'll learn it if it is useful to them. If not your only going to alienate some of the user base by making it difficult for them not to.

Perfect solution for me would be x to close. Hover over x to see keyboard alternative.

jeffmcneill:
Aggree with @reggieb the lack of a visual cue (x), and needing to read instructions or remember a keyboard command (esc) is a step back in modern usability. http://www.nngroup.com/articles/recognition-and-recall/

sotomo:
Esc is fine with me.  Rouge X buttons confuse me.  

IMO, the larger problem is that the search pane is not really a window or pane (that can be moved around).  I recommend making the search area a window or pane, then an X button would make more sense.

jeremykoerber:
I had this problem for a while and nothing was working. It turned out to be an issue with the git-diff-details package. I disabled it and I was able to close the find panel with ESC again. Found the issue by using the toggling the keybinding resolver and hitting ESC.

ea167:
I'm new to Atom, and ESC is NOT intuitive AT ALL. 
I had to google it after deep frustration. Not good for such a basic feature as Find.

You should definitely go with a X button, and tooltip the "Esc" when hovering

ojwoodford:
X button to close the find panel, please. :+1:

DaClone:
I'm also new to Atom and found the lack of any obvious way to close the search pane.
I, like some other people here, had to Google it.

A X button (or really any button at all) to close it with a tooltip for ESC would solve it easily.

Vloz:
As said, that's 100% not intuitive.

FoolishFrost:
Jast had to search to figure this out as well, and I'm a mac/pc/linux programmer that goes back to the C64 and vic20 era.

[...]

PR #687, opened 2016-03-28
Add a close button ( #347 ) to the find and find in project panels

megametres:
Hello, a simple change to add a close button #347

I dont think that the "Close this panel with the esc key" message is a viable alternative to the close button.

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -25,11 +25,11 @@ class FindView extends View
 
     @div tabIndex: -1, class: 'find-and-replace', =>
       @header class: 'header', =>
-        @span outlet: 'descriptionLabel', class: 'header-item description', 'Find in Current Buffer'
-        @a outlet: 'closeButton', class: 'icon icon-x pull-right octicons'
-        @span class: 'header-item options-label pull-right', =>
-          @span 'Finding with Options: '
+        @span outlet: 'descriptionLabel', class: 'header-item description inline-block', 'Find in Current Buffer'
+        @span class: 'header-item options-label', =>
+          @span 'With Options: '
           @span outlet: 'optionsLabel', class: 'options'
+        @a outlet: 'closeButton', class: 'icon icon-x pull-right octicons'
 
       @section class: 'input-block find-container', =>
         @div class: 'input-block-item input-block-item--flex editor-container', =>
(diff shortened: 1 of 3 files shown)


simurai:
I'm flip-flopping on this.

- Adding a close button is common and would help initially.
- Adding a close button would train people to use it, while there is an easier way.

Here the issue where the "Close this panel with the `esc` key" got added instead of a close button: https://github.com/atom/find-and-replace/issues/164

/cc @atom/feedback Any strong opinions about it?

thomasjo:
For me this boils down to *affordance* (in the UX sense) — while we can train people to press esc, it is not truly intuitive, and also not discoverable. Hence poor affordance. The affordance of a close button on the other hand, is much better. It affords being pressed, and is immediately discoverable and intuitive.

**TL; DR:** I vote :+1: on this.

thomasjo:
Also if we add the close button, we can teach people about the esc shortcut via a tooltip on the close button.

simurai:
*If* the close button gets added, having the "Close this panel with the `esc` key" feels strange on the left. Also, the `x` could be confused with clearing the options and not closing the panel.

screen shot 2016-03-29 at 2 48 40 pm

- So yeah, maybe the "Close this panel with the `esc` key" with a tooltip on the `x`
- Center align the options

screen shot 2016-03-29 at 3 44 49 pm

megametres:
I have another concern about the same topic, I'd like to change the shortcut ctrl+F to be on "Toggle Find in buffer" instead of "Find in buffer" and the same for ctrl+shift+F

It is also related to the closing of the panel, do anyone knows why it's not already this way and if there are any objections on the matter?

50Wliu:
> do anyone knows why it's not already this way

<kbd>Ctrl+F</kbd> is meant to _re-focus_ the panel.  Imagine the following scenario: You open <kbd>Ctrl+F</kbd> to search for something and find a result that you need to change, so you edit the text.  Then instead of having to manually click in the find box you can just <kbd>Ctrl+F</kbd> again to bring back focus to the find panel, and keep on searching.

megametres:
I made the suggested adjustment with one detail, I find more useful/pretty the "Finding with option" aligned to the left
close-button

elharo:
+1

simurai:
Made a PR to move the labels back to the right: https://github.com/megametres/find-and-replace/pull/1

I guess this PR also needs a spec?

maxbrunsfeld:
@megametres Thanks for your work on this. I'm going to close this out until you get a chance to address the remaining issues. FYI, we're planning on making some big changes to this package in order to stop using jQuery, so further conflicts may arise.

PR #697, opened 2016-04-06
Remove backwards-compatible change now that MarkerLayers shipped

as-cii:
This allows us to avoid using custom properties when marking ranges, which is going to be deprecated with the DisplayLayer refactoring. This should be a safe change now that MarkerLayers have been released on both Atom stable and beta.

/cc: @nathansobo


nathansobo:
👍

PR #702, opened 2016-04-08
Ability to see context of project search results - More lines

### LINKED ISSUE #190: Extra lines of context above/below search results
ST includes two lines above and below search results for context, which I regularly found useful and would love to see included in Atom.

Ideally the line count would be a setting instead of hardcoded:

```
[ ] Focus Editor After Search
[ ] Open Project Find Results In Right Pane
[2] Lines of Context Above/Below Search Results
```

:beers:

MartinMuzatko:
Closes #190 
This is my first contribution to an official core atom package.

I know that this PR is missing the ability to configure the amount of lines to see.
Any hints on how to improve this PR is very much appreciated.

I acknowledge the importance of this package to work properly, so please point out any required improvements.

The additions are not tested via Jasmine/Mocha or whatever testing engine you use. No specs provided.

**Features**
- [x] Show extra lines of context
- [ ] Correct indentation
- [ ] Overlapping lines should resolve to only show what is required
- [ ] Gaps between code in the same file should be expandable (max gap: 3 lines?)
- [x] Set amount of lines to see via config
- [ ] Results on the same line or context should appear on that result
- [ ] Remove empty lines in result (maybe)
- [ ] Syntax highlighting (maybe)

**Fixing known issues**
- [ ] Arrow Key navigation
- [ ] Even if there are results, the statusbars say "No results for ..."
- [x] Only buildBuffer or save lines that are required within the model

**Current Status**
image

### DIFF
--- a/lib/project/results-model.coffee
+++ b/lib/project/results-model.coffee
@@ -210,10 +210,12 @@ class ResultsModel
       filePathInsertedIndex = binaryIndex(@paths, filePath, stringCompare)
       @paths.splice(filePathInsertedIndex, 0, filePath)
 
-    @matchCount += result.matches.length
-
-    @results[filePath] = result
-    @emitter.emit 'did-add-result', {filePath, result, filePathInsertedIndex}
+    bufferPromise = atom.project.buildBuffer(filePath)
+    bufferPromise.then (buffer) =>
+      result.lines = buffer.lines
+      @matchCount += result.matches.length
+      @results[filePath] = result
+      @emitter.emit 'did-add-result', {filePath, result, filePathInsertedIndex}
 
   removeResult: (filePath) ->
     if @results[filePath]
--- a/lib/project/results-model.coffee
+++ b/lib/project/results-model.coffee
@@ -1,5 +1,5 @@
 _ = require 'underscore-plus'
-{Emitter} = require 'atom'
+{Emitter, Range} = require 'atom'
 escapeHelper = require '../escape-helper'
 
 class Result
@@ -212,7 +212,9 @@ class ResultsModel
 
     bufferPromise = atom.project.buildBuffer(filePath)
     bufferPromise.then (buffer) =>
-      result.lines = buffer.lines
+      for match in result.matches
+        range = Range.fromObject(match.range)
+        match.lines = buffer.lines.slice(range.start.row, range.start.row+2+1+2)
       @matchCount += result.matches.length
       @results[filePath] = result
       @emitter.emit 'did-add-result', {filePath, result, filePathInsertedIndex}
--- a/lib/project/results-model.coffee
+++ b/lib/project/results-model.coffee
@@ -215,7 +215,7 @@ class ResultsModel
       for match in result.matches
         range = Range.fromObject(match.range)
         linesAmount = atom.config.get('find-and-replace.searchContextExtraLines')
-        match.lines = buffer.lines.slice(range.start.row, range.start.row+linesAmount*2+1)
+        match.lines = buffer.lines.slice(range.start.row-linesAmount, range.end.row+linesAmount+1)
       @matchCount += result.matches.length
       @results[filePath] = result
       @emitter.emit 'did-add-result', {filePath, result, filePathInsertedIndex}
--- a/lib/project/results-model.coffee
+++ b/lib/project/results-model.coffee
@@ -214,7 +214,8 @@ class ResultsModel
     bufferPromise.then (buffer) =>
       for match in result.matches
         range = Range.fromObject(match.range)
-        match.lines = buffer.lines.slice(range.start.row, range.start.row+2+1+2)
+        linesAmount = atom.config.get('find-and-replace.searchContextExtraLines')
+        match.lines = buffer.lines.slice(range.start.row, range.start.row+linesAmount*2+1)
       @matchCount += result.matches.length
       @results[filePath] = result
       @emitter.emit 'did-add-result', {filePath, result, filePathInsertedIndex}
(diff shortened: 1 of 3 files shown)


DouweM:
@MartinMuzatko Can you add some screenshots?

And a few questions:
- What happens with overlap? Say we have results on line 12 and 15. We would expect to see 10..17, without 13 and 14 repeated
- What happens with gaps? Say we have results on line 8 and 15. We would expect to see 6..10 and 13..17. Between those, it would be nice to have a line with `..`, like Sublime does, see https://github.com/atom/find-and-replace/issues/190#issuecomment-120517904.

MartinMuzatko:
@DouweM that is true, there is nothing implemented to detect line collisions as shown below:

image

Also what bugs me, is that the matches are not separated. it is just a blog of code with the highlight in the middle.

The `..` feature is also missing.

I'll continue working on this. Thank you very much for your input @DouweM !

DouweM:
> Is there a way to detect tabs? I do have the space in the lines, but I have to find out what other options are in the buffer class. Maybe also the identation

This is likely simply a CSS issue where consecutive spaces are collapsed into one, note that in your screenshot, line 35 appears to have one space of indentation even though it should have more. Try `white-space: pre;`

> Also what bugs me, is that the matches are not separated. it is just a blog of code with the highlight in the middle.

I don't mind that, actually. It's the same way Sublime did it: https://github.com/atom/find-and-replace/issues/190#issuecomment-120517904

DouweM:
@MartinMuzatko Another suggestion while you're at this: Have the search results use the same monospace font as elsewhere in text editors.

MartinMuzatko:
Now I need to deal with Promises, because I need to build the buffer in order to get the lines:

`atom.project.buildBuffer(filePath)`

MartinMuzatko:
Currently I'm learning how to deal with js scope in coffeescript. Any help to solve the promise problem is very much appreciated.

```coffeescript
class MatchView extends View
  @content: (model, {filePath, match}) =>
    linesPromise = atom.project.buildBuffer(filePath)
    linesPromise.then (lines) ->
      @getContent(model, {filePath, match, lines})

  getContent: (@model, {filePath, match}) ->
    range = Range.fromObject(match.range)
    matchStart = range.start.column - match.lineTextOffset
    ......
```
It seems like the scope of the promise has no access to @ (this)

DouweM:
@MartinMuzatko Check out the fat arrow: http://coffeescript.org/#fat-arrow

MartinMuzatko:
I have another problem of finding documentation on the View Class. It seems like I have to find another way around, the fat arrow still gives me the window as scope..

I think the best thing to do, is to actually save suffix/prefix in the result as the search is done, not when the view is output. There I can retrieve all the data before I return a view. 

image

DouweM:
@MartinMuzatko Your `content` method is a class method, so `@` will refer to `MatchView` (note that `this` != `@` when using `=>`). Your `getContent` is an instance method, before you can use `@getContent`, you need to instantiate an instance of `MatchView` and work in the context of that.

MartinMuzatko:
Thank you for the pointers, I'll do my best to fix this :)

MartinMuzatko:
Have another look please. The navigation doesn't work that great yet, buffer lines are now retrieved via model, where it belongs to.
I will soon work to implement the **features**, namely:

- [ ] Correct indentation
- [ ] Overlapping lines should resolve to only show what is required
- [ ] Gaps between code in the same file should be expandable (max gap: 3 lines?)

**Fixing known issues**
- [ ] Arrow Key navigation

lib/project/results-model.coffee
```diff
@@ -210,10 +210,12 @@ class ResultsModel
       filePathInsertedIndex = binaryIndex(@paths, filePath, stringCompare)
       @paths.splice(filePathInsertedIndex, 0, filePath)
 
-    @matchCount += result.matches.length
-
-    @results[filePath] = result
-    @emitter.emit 'did-add-result', {filePath, result, filePathInsertedIndex}
+    bufferPromise = atom.project.buildBuffer(filePath)
+    bufferPromise.then (buffer) =>
+      result.lines = buffer.lines
```
DouweM: I'm not sure if it's a good idea to pass all lines rather than just those that will be used.

I'm not on the Atom team or anything though, so they may be fine with it :)

MartinMuzatko: I'm not sure if I can tell `buildBuffer` just to build the lines I need. The API Docs are really missing a lot of information unfortunately: https://atom.io/docs/api/v1.6.2/Project.

However, from the modelside, I can just use the lines required :)

MartinMuzatko:
Recently continued my work, commits are here soon.
I'm getting used to coffeescript, but I ran into another problem:

image

Should we just merge resutls that are in the same line? like

> `lines` = match.`lines`

DouweM:
@MartinMuzatko That would be really nice

wonderbeyond:
How is the progress? I want the feature.

MartinMuzatko:
Hello there! Currently I'm pretty busy. The results so far is not quite done yet. Please feel free to build upon what I have created so far. I'll resume once I have some time again for open source projects.

Thanks

wonderbeyond:
@MartinMuzatko You are great!
I cannot help more since I'm unfamiliar with Atom internal.
I've followed you at GitHub and Twitter.

cvogt:
+1

wonderbeyond:
I want you! @MartinMuzatko

dirk-thomas:
Please see #847 for another working proposal.

dirk-thomas:
With #847 being merged this can probably be closed.

PR #704, opened 2016-04-13
Always open search results on single click

BinaryMuse:
This was a nice idea but felt too unintuitive.


BinaryMuse:
This fails against 1.6.2 but passes against 1.7.0

Issue #634, opened 2016-01-05
Displaying deleted files


adambuczynski:
I assume find & replace uses some kind of caching system for all files when searching in project files, but when I delete a file through Atom, it should probably trigger an event to remove that file from the cache.

Currently, after having deleted a file, it will still show up in the search results, which is confusing at best. 

Possibly related, but the fuzzy finder module is also plagued by this issue, see https://github.com/atom/fuzzy-finder/issues/168

adambuczynski:
It may have to do with the file still being open in a tab (see https://github.com/atom/tabs/issues/160)
To reproduce:

1. Create a test file, add some easy to find content like `ABCABCABC`
2. Save the file
3. Verify that searching for `ABCABCABC` in files shows up the file
4. Close search panel
5. Delete the file from tree view
6. Search in files again for `ABCABCABC`
7. The deleted file shows up, clicking it opens a new tab with empty contents:

![image](https://cloud.githubusercontent.com/assets/490562/12934869/9abd9db6-cff6-11e5-887e-9b4e7b83f8a4.png)

acontreras89:
@anderoonies can you test if this is (as well) due to `node-pathwatcher` not firing `delete` events for files moved to the trash on OSX?

anderoonies:
gladly, i'll update as soon as i get back to my machine and have a chance

acontreras89:
atom/node-pathwatcher#106 has been merged, which may solve this problem. @adambuczynski can you ensure this is working properly with the latest version of `node-pathwatcher`?

adambuczynski:
@acontreras89 so this is partially fixed, I guess it depends on how find and replace should work.

When I follow the above steps now, the tab of the deleted file is properly removed, and when I close the find and replace panel and open it again to search, the deleted file does not show up.

However, when you leave the panel open and try searching for the same term again, it does not perform a new search, and your previous result (including the deleted file) still shows up. This may or may not be the intended behaviour of find and replace, so I can't classify it as an outright bug. But when you click on the deleted file in the search results, the same happens, e.g. an empty tab is opened with the same file name, but pointing to a non existing file. After you do that, the result disappears from the find and replace search results.

lee-dohm:
Personally, I think it would be best if the corresponding entries could be removed from the find-in-project view.

acontreras89:
I created a pull request to address this. Will wait for feedback before working any longer on it, though.

PR #705, opened 2016-04-16
Watch paths to remove results on file deletion

acontreras89:
**WORK IN PROGRESS** 

Fixes https://github.com/atom/find-and-replace/issues/634.

Current specs are failing when adding results manually ([like here](https://github.com/atom/find-and-replace/blob/master/spec/results-view-spec.coffee#L69)), as `node-patwatcher` is unable to watch the path.
Specs for the introduced change are missing.

### DIFF
--- a/lib/project/results-model.coffee
+++ b/lib/project/results-model.coffee
@@ -72,6 +72,7 @@ class ResultsModel
     @matchCount = 0
     @regex = null
     @results = {}
+    @watchSubscriptions[path].close() for path of @watchSubscriptions
     @watchSubscriptions = {}
     @paths = []
     @active = false
--- a/lib/project/results-model.coffee
+++ b/lib/project/results-model.coffee
@@ -72,6 +72,7 @@ class ResultsModel
     @matchCount = 0
     @regex = null
     @results = {}
+    @watchSubscriptions = {}
     @paths = []
     @active = false
     @searchErrors = null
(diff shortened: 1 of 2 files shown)


lee-dohm:
This looks pretty good. You may want to wrap the pathwatcher stuff in `Disposables` and `CompositeDisposables`:

https://atom.io/docs/api/v1.7.1/Disposable#instance-constructor

This would make cleaning up at the end much easier.

acontreras89:
Hey @lee-dohm, thanks for your comments.
I see how I could encapsulate each _watcher_ inside a `Disposable`. However, since I need to close them individually at some points (i.e. when a file is deleted,) I don't see how I could be using a `CompositeDisposable` to aggregate them. Can you point me in the right direction?

Oh, and I'll be updating existing specs and adding the new ones soon.

lee-dohm:
@acontreras89 Yep, you're right :grinning:

maxbrunsfeld:
@acontreras89 I'm going to close this one out for the time being, until you get a chance to come back to it. It's a cool idea. FYI, we're planning on making some big changes to this package soon, to remove jQuery, so further conflicts may be created.

acontreras89:
Hello @maxbrunsfeld
I'd like to finish this, but I haven't contributed to atom in a while, so I'm feeling a bit rusty. Can you point me in the right direction to (1) find a way around current specs failing and (2) adding tests for this?

Issue #92, opened 2013-12-05
Autocomplete with words from buffer


kevinsawicki:
In the Find/Replace editors.

See atom/atom#519

andrewpcone:
Would it be appropriate to use `C-space` for this purpose, like in the main buffer? Or does this clobber something I'm not realizing?

PR #710, opened 2016-05-06
Enable autocompletion in the find editor

wvanlint:
This PR enables autocompletion in the find editor by:
1. Registering the find editor so it's observable by `autocomplete-plus`
2. Adding a corresponding CSS selector to `SymbolProvider` so it knows to provide completions for that editor.

Fixes atom/find-and-replace#92

/cc @as-cii

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -14,6 +14,9 @@ class FindView extends View
       softWrapped: false
       buffer: findBuffer
       placeholderText: 'Find in current buffer'
+    if atom.textEditors?
+      disposable = atom.textEditors.add(findEditor)
+      findEditor.onDidDestroy -> disposable.dispose()
 
     replaceEditor = buildTextEditor
       mini: true
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -115,6 +115,16 @@ module.exports =
       'find-and-replace:select-skip': (event) ->
         selectNextObjectForEditorElement(this).skipCurrentSelection()
 
+    # Request autocompletions for the find field.
+    autocompletePackage = atom.packages.activePackages['autocomplete-plus']
+    if autocompletePackage?
+      autocompleteManager = autocompletePackage.mainModule.autocompleteManager
+      for providerData in autocompleteManager.providerManager.providers
+        if providerData.provider.constructor.name is 'SymbolProvider'
+          providerData.provider.addTextEditorSelector?(
+            '.find-container atom-text-editor')
+          break
+
   provideService: ->
     resultsMarkerLayerForTextEditor: @findModel.resultsMarkerLayerForTextEditor.bind(@findModel)
 


lib/find.coffee
```diff
@@ -115,6 +115,16 @@ module.exports =
       'find-and-replace:select-skip': (event) ->
         selectNextObjectForEditorElement(this).skipCurrentSelection()
 
+    # Request autocompletions for the find field.
+    autocompletePackage = atom.packages.activePackages['autocomplete-plus']
+    if autocompletePackage?
+      autocompleteManager = autocompletePackage.mainModule.autocompleteManager
+      for providerData in autocompleteManager.providerManager.providers
+        if providerData.provider.constructor.name is 'SymbolProvider'
```
wvanlint: This is a bit ugly, but I think it's okay for now. In the future, providers could ideally be registered together with a name.
It would also allow you to register multiple instances of the same provider with different options e.g. an `ActiveBufferSymbolProvider`, `AllBuffersSymbolProvider`, `ProjectSymbolProvider`, etc.

lib/find-view.coffee
```diff
@@ -14,6 +14,9 @@ class FindView extends View
       softWrapped: false
       buffer: findBuffer
       placeholderText: 'Find in current buffer'
+    if atom.textEditors?
+      disposable = atom.textEditors.add(findEditor)
+      findEditor.onDidDestroy -> disposable.dispose()
```
as-cii: What do you think about disposing the `Disposable` object that `onDidDestroy` returns, when deactivating the package?

lib/find.coffee
as-cii: Seems like this could be problematic if the autocomplete-plus package gets activated *after* the find-and-replace package. It also seems like `SymbolProvider` gets re-initialized every time the user toggles the `autocomplete-plus.enableBuiltinProvider`, thus disabling autocompletion until the next reload.

I wonder if we could provide a service in ac+ instead and let it expose an interface like the following:

```coffee
{
  getDefaultProvider: ->
    @defaultProvider

  observeProviders: (callback) ->
    callback(provider) for provider in @providers
    @emitter.on 'did-register-provider', callback
    
  onDidRemoveProvider: (callback) -> 
    @emitter.on 'did-remove-provider', callback
}
```

We might add even more methods to it, but I am a little unclear on what are the expansion points we may want to provide, so I think it's okay to keep the API tight so that we can support just this use case for now, while still potentially allowing package authors to use it too. After we do that, we could use the service like this:

```coffee
consumeAutocompletePlus: (autocompletePlus) ->
  this.subscriptions.add autocompletePlus.observeProviders (provider) ->
    if autocompletePlus.getDefaultProvider() is provider
      autocompletePlus.getDefaultProvider().addTextEditorSelector?('.find-container atom-text-editor') 
```

This would also prevent the details of the autocomplete-plus' implementation to leak into find-and-replace, because we wouldn't rely on the `SymbolProvider` magic constant anymore.

What do you think?

maxbrunsfeld:
@wvanlint Thanks for your work here. I'm going to close this out until you get a chance to come back to it. FYI, we're going to make big changes to this package soon in order to remove jQuery, so some additional conflicts will be created.

PR #719, opened 2016-05-11
Add ability to pull selected text into replace field

ruthgrace:
…t-e for slurpping selected text into replace field, as requested in #67

The hotkey is cmd-shift-e for Mac, and ctrl-shift-e for Windows/Linux. If this pull request and pull request #713 are accepted, then issue #67 should be resolved.

Please let me know if you have any feedback on this pull request. It's my first time writing specs! Thanks very much in advance.

### DIFF
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -50,13 +50,18 @@ module.exports =
     @subscriptions.add atom.commands.add 'atom-workspace', 'find-and-replace:use-selection-as-find-pattern', =>
       return if @projectFindPanel?.isVisible() or @findPanel?.isVisible()
 
+      @createViews()
+      @projectFindPanel.hide()
+      @findPanel.show()
+      @findView.focusFindEditor()
+
     @subscriptions.add atom.commands.add 'atom-workspace', 'find-and-replace:use-selection-as-replace-pattern', =>
       return if @projectFindPanel?.isVisible() or @findPanel?.isVisible()
 
       @createViews()
       @projectFindPanel.hide()
       @findPanel.show()
-      @findView.focusFindEditor()
+      @findView.focusReplaceEditor()
 
     @subscriptions.add atom.commands.add 'atom-workspace', 'find-and-replace:toggle', =>
       @createViews()
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -50,6 +50,9 @@ module.exports =
     @subscriptions.add atom.commands.add 'atom-workspace', 'find-and-replace:use-selection-as-find-pattern', =>
       return if @projectFindPanel?.isVisible() or @findPanel?.isVisible()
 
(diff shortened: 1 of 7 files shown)


lib/find.coffee
```diff
@@ -50,6 +50,9 @@ module.exports =
     @subscriptions.add atom.commands.add 'atom-workspace', 'find-and-replace:use-selection-as-find-pattern', =>
       return if @projectFindPanel?.isVisible() or @findPanel?.isVisible()
 
+    @subscriptions.add atom.commands.add 'atom-workspace', 'find-and-replace:use-selection-as-replace-pattern', =>
```
50Wliu: It looks like adding this broke the `use-selection-as-find-pattern` above as now it doesn't have any actual code associated with it.

ruthgrace: The use-selection-as-find-pattern specs still all pass though. Should this code read like this instead?
```
  handleEventsForSelectionAsReplace: ->
    @subscriptions.add atom.commands.add 'atom-workspace',
      'find-and-replace:use-selection-as-replace-pattern': @setSelectionAsReplacePattern
```

ruthgrace:
My first commit broke the functionality where if `meta-e` (for use-selection-as-find-pattern) was pressed before the `projectFindView` and `findView` existed, the views would be created, and focus would be put on the find editor. I've fixed this with a second commit, however, I would like the views to be created as described also when the use-selection-as-replace-pattern function is called first, and this is not happening right now. What would be preventing the views from being created in [these lines of code](https://github.com/ruthgrace/find-and-replace/blob/b5d1f73b6d325d9fc5095b28999a7b434a3085a6/lib/find.coffee#L61-L64)?

ruthgrace:
50Wliu helped me figure it out! I had forgotten to add the new use-selection-as-replace-pattern to the activation commands. Now if `cmd-shift-e` or `ctrl-shift-e` is pressed right after atom starts up, the find-and-replace package activates, the find panel becomes visible, and focus is put on the replace editor. Everything now works as desired.

lib/find.coffee
```diff
@@ -55,6 +55,14 @@ module.exports =
       @findPanel.show()
       @findView.focusFindEditor()
 
+    @subscriptions.add atom.commands.add 'atom-workspace', 'find-and-replace:use-selection-as-replace-pattern', =>
+      return if @projectFindPanel?.isVisible() or @findPanel?.isVisible()
+
+      @createViews()
+      @projectFindPanel.hide()
```
50Wliu: Can we use showPanel to replace this line and the two below it?

ruthgrace: Sure. Should we do the same replacement for the
use-selection-as-find-pattern code above too?

ruthgrace: (In a separate PR?)

50Wliu: Yeah, that sounds good.

PR #721, opened 2016-05-14
Add an "Instant Search" setting that removes the 300ms delay

xob:
I'm sure there's a way to avoid having to restart Atom after the Instant Search setting is toggled, but I don't know how to do that. I'm open to any and all suggestions.

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -107,6 +107,7 @@ class FindView extends View
   destroy: ->
     @subscriptions?.dispose()
     @tooltipSubscriptions?.dispose()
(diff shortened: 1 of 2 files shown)


nixel2007:
@xob you need to add a subscription to "observe" the setting.

See - https://github.com/xDrivenDevelopment/atom-language-1c-bsl/blob/master/lib/language-1c-bsl.coffee#L14

maxbrunsfeld:
In the coming months, we're going to make find and replace faster and remove the 300ms delay entirely. Until then, I'd prefer not to add a setting for this. Thanks anyway!

PR #736, opened 2016-05-28
Enable and disable search in selection automatically

yufanyufan:
The objective of this enhancement is to automatically activate search within selection when find-and-replace is called and the selected text has one or more '/n' if no newline is found the selected text is used as the search pattern as it was implemented before. 

This is a reimplement of PR #616

This is how most editor works if the editor supports search in selection. For example, this is exactly how Visual Studio Code works.

I want to also point how it can be very tricky to turn on and off "search in selection" manually.
For example, if you select a word, and then ctrl-f, but forget to turn off "search in selection", then atom reports only one match, which is extremely misleading.

If you still have concern, I can provide a config option to turn this function on.

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -205,11 +205,23 @@ class FindView extends View
       'find-and-replace:replace-next': @replaceNext
       'find-and-replace:replace-all': @replaceAll
 
+  activateSelectionOption: =>
+    # This forces search the selection again if selection is changed.
+    @search(inCurrentSelection: false)
+    @search(inCurrentSelection: true)
+
+  deactivateSelectionOption: =>
+    @search(inCurrentSelection: false)
+
   focusFindEditor: =>
     selectedText = atom.workspace.getActiveTextEditor()?.getSelectedText?()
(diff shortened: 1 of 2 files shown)


hgoodman:
It would be nice to see this implemented in the core package. There is [an extension](https://github.com/Kraoz/atom-auto-replace-in-selection) that supposedly does this but it doesn't work and is apparently not maintained.

lawrence-dol:
@hgoodman Actually, it sort of does. If you hit CTRL-F repeatedly, it toggles the option, and if the last thing you did was not a find/replace then it does seem to default correctly. So it's slightly better than nothing.

That said, this pull request is what I want. I wish that PR's didn't languish like this in Atom projects; submitted in 2016 and no-one on the project team can resolve the conflicts?

lee-dohm:
Thanks for the interest and the work you put in on implementing this.

We discussed this as a team and we agree that [my original concerns](https://github.com/atom/find-and-replace/pull/616#issuecomment-164096138) about this feature being hard to discover and confusing still apply. Because of this, we won't be adding this feature.

Thanks for your interest in Atom.

Issue #615, opened 2015-12-08
Odd behavior when replacing with backslashes


Ingramz:
When doing a search on string `a b a b a b a`, searching for `b` and then replacing all with regular expression `\\\\\\\\` (four escaped backslashes), we get `a \\\\ a \\ a \ a` - first occurrence gets the desired four backslashes, the next one only two, and the last one a single one.

If there were more `b`-s, these would get also a single backslash.

The last build I have, is with find-and-replace version 0.191.0

Ingramz:
![untitled](https://cloud.githubusercontent.com/assets/1017132/11672168/d64c1c98-9e16-11e5-924f-f84a1ce37147.gif)

50Wliu:
Fixed by #740.

PR #740, opened 2016-06-08
Unescape escape sequence once on replacing buffer using regex

### LINKED ISSUE #739: "\\n" in replace string generates line ending after second match
Before "Replace All":
screenshot 2016-06-08 13 16 33
After "Replace All":
screenshot 2016-06-08 13 16 43

norio-nomura:
Fix #739

### DIFF
--- a/lib/buffer-search.coffee
+++ b/lib/buffer-search.coffee
@@ -70,11 +70,11 @@ class BufferSearch
     @findOptions.set({replacePattern})
 
     @editor.transact =>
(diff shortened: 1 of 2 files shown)


50Wliu:
Thanks for this @norio-nomura - can you please also add some specs so that we can make sure this behavior doesn't regress?

norio-nomura:
Added spec. 😃

50Wliu:
Also fixes #615.

norio-nomura:
Thanks! :pray:

PR #759, opened 2016-07-22
Add feature select-all-under

agleis:
- Selects all words matching the selected word from the cursor to the end of file
- Selects full words if select-next (ctrl-d, cmd-d) was used, and will match subwords as well if normal cursor selection was used.
- Similar to select-all, but will not 'wrap' back to the top of the file

ref: atom#67

select-all-under


50Wliu:
In http://docs.sublimetext.info/en/latest/reference/commands.html, find_all_under is described as "Finds all occurrences of the current selection or the current word.", which seems a lot like find_all to me.  Maybe  find_all was renamed to find_all_under at some point?

agleis:
Hm, I didn't realize that. In that case, this may be new functionality - finding all occurrences of the current selection from the cursor down to the end of the file. Sorry for not reading the Sublime docs too closely. But this still does seem useful to me, so maybe it could still work?

maxbrunsfeld:
@agleis It looks like you did a nice job implementing this, but I think this command is a bit too specific to be included in this core package. Thanks for diving into the code anyway!

PR #762, opened 2016-07-24
Reset search query when initialised

Alhadis:
This PR fixes a rather esoteric bug that blocks a user from rerunning the session's last buffer search.

## Steps to reproduce
1. Open a project file in Atom, and search for something that will match
2. Close project
3. Reopen project and project's file
4. Bring up the buffer search and enter the last search term verbatim. This should be done quickly, either by typing fast or pressing the <kbd>Up Arrow</kbd> key to access search history.
5. Hit <kbd>Enter</kbd>

**Expectation:** Markers should appear in the buffer
**Reality:** The program beeps at you as if the field was left blank

### DIFF
--- a/lib/find-options.coffee
+++ b/lib/find-options.coffee
@@ -16,7 +16,7 @@ class FindOptions
   constructor: (state={}) ->
     @emitter = new Emitter
 
-    @findPattern = state.findPattern ? ''
+    @findPattern = ''
     @replacePattern = state.replacePattern ? ''
     @pathsPattern = state.pathsPattern ? ''
     @useRegex = state.useRegex ? atom.config.get('find-and-replace.useRegex') ? false
--- a/spec/buffer-search-spec.coffee
+++ b/spec/buffer-search-spec.coffee
@@ -21,7 +21,7 @@ describe "BufferSearch", ->
       -----------
     """
 
-    findOptions = new FindOptions
+    findOptions = new FindOptions findPattern: "a+"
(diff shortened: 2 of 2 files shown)


50Wliu:
I can confirm both that the bug exists and that this PR fixes said bug.  However, looking at replacePattern and pathsPattern, does the bug manifest there as well?

Alhadis:
Nope. Those properties [aren't checked](https://github.com/atom/find-and-replace/blob/master/lib/buffer-search.coffee#L20-L25) in BufferSearch's `onDidChange` callback. I also tested this to see if they *did* need changing, which they didn't.

50Wliu:
Not familiar with this code, so please bear with me.

Then can we simply remove `changedParams.findPattern?` from that onDidChange block?  Because it looks like everything else there are actual *options*.

Alhadis:
Hrm yeah, you're right, actually. [From a glance](https://github.com/atom/find-and-replace/commit/ef0f673dfc72715546b26c14d20a326464025a54), it would seem there was a bit of copy+paste going on, so I'd say that check can be dropped. :)

Alhadis:
Oh! No, wait, it's needed because it represents the current state.

It should probably be called `SearchState` or something like that...

50Wliu:
Can you clarify on what the "state" is?

Alhadis:
Well for lack of a better word, I mean "state" as in "the state of the user's search criteria". E.g., the find-field, replace-field, and the options affecting their usage.

PR #768, opened 2016-08-11
Add feature collapse and expand results

jccguimaraes:
I often find myself scrolling a big list of results just to find a specific file. Following the `styleguide` approach. I made this small UI change.

collapse-expand

### DIFF
--- a/lib/project/results-pane.coffee
+++ b/lib/project/results-pane.coffee
@@ -14,8 +14,7 @@ class ResultsPaneView extends ScrollView
         @span outlet: 'previewCount', class: 'preview-count inline-block'
         @div outlet: 'previewControls', class: 'preview-controls', =>
           @div class: 'btn-group', =>
-            @button outlet: 'collapseAll', class: 'btn'
-            @button outlet: 'expandAll', class: 'btn'
+            @button outlet: 'toggleExpanded', class: 'btn'
         @div outlet: 'loadingMessage', class: 'inline-block', =>
           @div class: 'loading loading-spinner-tiny inline-block'
           @div outlet: 'searchedCountBlock', class: 'inline-block', =>
@@ -36,12 +35,9 @@ class ResultsPaneView extends ScrollView
     @on 'focus', @focused
 
     @previewControls.hide()
-    @collapseAll
-      .text('Collapse All')
-      .click(@collapseAllResults)
-    @expandAll
-      .text('Expand All')
-      .click(@expandAllResults)
+    @toggleExpanded
+      .text('Toggle Expanded')
+      .click(@toggleExpandedResults)
 
   attached: ->
     @model.setActive(true)
@@ -153,10 +149,6 @@ class ResultsPaneView extends ScrollView
       @previewControls.hide()
       @addClass('no-results')
 
-  collapseAllResults: =>
-    @resultsView.collapseAllResults()
-    @resultsView.focus()
-
-  expandAllResults: =>
-    @resultsView.expandAllResults()
+  toggleExpandedResults: =>
+    @resultsView.toggleExpandedResults()
     @resultsView.focus()
--- a/lib/project/results-pane.coffee
+++ b/lib/project/results-pane.coffee
@@ -14,7 +14,8 @@ class ResultsPaneView extends ScrollView
         @span outlet: 'previewCount', class: 'preview-count inline-block'
         @div outlet: 'previewControls', class: 'preview-controls', =>
           @div class: 'btn-group', =>
-            @button outlet: 'toggleExpanded', class: 'btn'
+            @button outlet: 'collapseAll', class: 'btn'
+            @button outlet: 'expandAll', class: 'btn'
         @div outlet: 'loadingMessage', class: 'inline-block', =>
           @div class: 'loading loading-spinner-tiny inline-block'
           @div outlet: 'searchedCountBlock', class: 'inline-block', =>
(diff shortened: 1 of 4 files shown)


lib/project/results-pane.coffee
```diff
@@ -140,3 +149,19 @@ class ResultsPaneView extends ScrollView
       @removeClass('no-results')
     else
       @addClass('no-results')
+
+  collapseAllResults: =>
+    i=0
```
50Wliu: I believe the style used in the rest of the codebase is `i = 0`

50Wliu:
This will also need some specs.

lib/project/results-pane.coffee
jccguimaraes: will make changes. Also, does it make sense to instead making the logic here or extend the `collapseResult` and `expandResult` in the `results-view.coffee`?

jccguimaraes:
I just seen it broke in appveyor 😢 damn

lib/project/results-pane.coffee
50Wliu: I took a look and I feel like these would belong more in `results-view`.  Definitely take my advice with a grain of salt though - I've never really touched find-and-replace before.

lib/project/results-pane.coffee
```diff
@@ -10,8 +10,12 @@ class ResultsPaneView extends ScrollView
 
   @content: ->
     @div class: 'preview-pane pane-item', tabindex: -1, =>
-      @div class: 'panel-heading', =>
+      @header class: 'preview-header', =>
         @span outlet: 'previewCount', class: 'preview-count inline-block'
+        @div class: 'preview-controls', =>
+          @div class: 'btn-group', =>
+            @button outlet: 'previewCollapse', class: 'btn'
```
50Wliu: Would it make more sense to call these `collapseAll` and `expandAll`?

jccguimaraes: yep

lib/project/results-pane.coffee
jccguimaraes: will change

jccguimaraes:
@50Wliu, could you help me out on this fails in AppVeyor beta channel? I can't replicate this issues 😢 😞

50Wliu:
Sure - I'll check this branch out and see if I can repro them.

50Wliu:
Ok.  I am getting the same failures when running tests locally (Atom 1.11.0-dev-67410f3), but I also can't replicate them.  Something I did notice though is that the border between the "header" and the results seem to have gone missing in this PR:

This PR:
no-border

master:
border

jccguimaraes:
@50Wliu finally did some changes that fixed all integrations. Also updated the gif on first entry. Can you please validate and if this is a runner?

Thanks

50Wliu:
To be honest, I'm not really liking the header UI changes, at least using One Light.  Obviously that could just be me so I'll defer to @simurai regarding that.

Regarding functionality, I did find a bug where the Collapse/Expand All button remains focused after pressing it.  I would expect that the focus returns to the actual results after clicking one of them.  This also breaks navigating using the keyboard.

jccguimaraes:
Regarding the first point i think i kept the current UI but I can check in more depth.

As of point two. Yeah that seems logic to implement.

Thanks for the review

jccguimaraes:
@50Wliu and @simurai any updates on this regarding the UI?

50Wliu:
My life has gotten about 1000% busier since my last comment.  Sorry about that and I'll try to get back to you when things are more under control.

styles/find-and-replace.less
```diff
@@ -228,7 +228,20 @@ atom-workspace.find-visible {
   display: flex;
   flex-direction: column;
 
-  .panel-heading {
+  .preview-header {
```
simurai: Not sure about the gradient and the top border, maybe that should be left for themes to add. Here where they would be removed:

```less
.preview-header {
    display: flex;
    padding: @component-padding/2 @component-padding/2 @component-padding/2 @component-padding;
    align-items: center;
    justify-content: space-between;
    font-weight: normal;
    border-bottom: 1px solid @panel-heading-border-color;
    background-color: @panel-heading-background-color;
  }
```

screen shot 2016-08-24 at 3 42 28 pm

simurai:
I think that's a nice change. 👍 

Something I wonder: Would it be easier to just have a single button that toggles between "Collapse All" and "Expand All". It's a little less flexible, but probably most of the time you just wanna collapse all and then open individual groups.

jccguimaraes:
@simurai i can leave with that :) was just extending the way it is in the `Styleguide` package.

jccguimaraes:
hi @simurai, do you have any update on this PR?
If it's a good feature to move on I need to validate against all the last updates, but if not, will drop this.

Thanks

simurai:
I would be in favor of just a single toggle button, but even with two, I think it's still a good improvement. I'll take another look tomorrow.

jccguimaraes:
@simurai thanks for the reply, i will transform it to a single button and double check everything.

jccguimaraes:
One of the reasons I did 2 buttons was because you can have individually entries expanded or collapsed. With one button only, there needs to be a logic to that:

1. A state variable `false`/`true`.
2. Starts expanded, then collapsed, and so on, with the exception that if all are in one state, it will do the opposition.

What do you thing @simurai?

50Wliu:
> One of the reasons I did 2 buttons was because you can have individually entries expanded or collapsed.

I agree with this and think the 2 buttons makes sense in this scenario.

jccguimaraes:
@50Wliu and @simurai, reverted to two buttons again. Please review if you can.

simurai:
> One of the reasons I did 2 buttons was because you can have individually entries expanded or collapsed.

I was thinking if you start to mix collapsed and un-collapsed and then you click the single button, if it doesn't do what you wanted, you just click a 2nd time. Not sure if that's less annoying than having to think which button to press.

Anyways, the 1 or 2 button discussion shouldn't hold up this PR any further. If somebody could review the code, otherwise 👍  to merge.

maxbrunsfeld:
:+1: The code and tests looks good to me.

PR #769, opened 2016-08-11
Add context to project search results

sullerandras:
Closes https://github.com/atom/find-and-replace/issues/190
Shows a few lines around the search results to provide context.
Similar to https://github.com/atom/find-and-replace/pull/702 which didn't work for me.

Current status:
image

Features:
- Set amount of lines to see via config (default: 0)
- Show extra lines of context
- Does not show overlapping lines
- Arrow Key navigation
- Shows '...' between contexts
- Shows first 140 chars of context lines, to avoid extensive CPU usage when displaying results in minified js files for example
- Fix indentation

Missing features:
- Results on the same line or context should appear on that result (maybe)
- Syntax highlighting (probably won't do this)

Haven't touched the specs so probably broke most of them.
Any feedback appreciated.


50Wliu:
> Results on the same line or context should appear on that result (maybe)

I would love this - multiple matches on the same line appearing as different results has always messed me up.

sullerandras:
> I would love this - multiple matches on the same line appearing as different results has always messed me up.

This is actually an independent issue of the find-and-replace package and probably needs quite a bit of refactoring to solve it. Although I agree that showing context lines makes it even more confusing.

lib/project/match-view.coffee
```diff
@@ -10,16 +10,33 @@ class MatchView extends View
...
 
-    @li class: 'search-result list-item', =>
-      @span range.start.row + 1, class: 'line-number text-subtle'
-      @span class: 'preview', outlet: 'preview', =>
-        @span prefix
-        @span match.matchText, class: 'match highlight-info', outlet: 'matchText'
-        @span match.matchText, class: 'replacement highlight-success', outlet: 'replacementText'
-        @span suffix
+    liStyle = if match.CONTEXT_LINES > 0 then 'padding-top: 0; padding-bottom: 0'
+    divStyle = 'white-space: pre'
```
50Wliu: Can this and the above style be placed in the `styles/` directory?  Maybe not `liStyle` because of the conditional but `divStyle` looks doable.

lib/project/result-view.coffee
```diff
@@ -44,11 +44,61 @@ class ResultView extends View
...
+    return if CONTEXT_LINES is 0
+    content = fs.readFileSync(filePath).toString()
+    lines = content.split('\n')
+    prevMatch = match
+    prevRowIndex = 0
+    prevLinesAfter = 0
+    for matchIndex in [0...matches.length]
+      match = matches[matchIndex]
+      range = match.range
+      if not range
```
50Wliu: This can be `unless range`

lib/project/result-view.coffee
```diff
@@ -44,11 +44,61 @@ class ResultView extends View
...
       for match in matches
         @matches.append(new MatchView(@model, {@filePath, match}))
 
     @matches.children().eq(selectedIndex).addClass('selected') if selectedIndex > -1
 
+  addContextToMatches: (filePath, matches) ->
+    CONTEXT_LINES = atom.config.get('find-and-replace.searchContextExtraLines')
+    return if CONTEXT_LINES is 0
+    content = fs.readFileSync(filePath).toString()
+    lines = content.split('\n')
```
50Wliu: I'm worried about Windows compatibility here - I assume this would still work because CRLF also includes the `\n`, but would the extra `\r` interfere with anything?

lib/project/match-view.coffee
sullerandras: Fixed in https://github.com/atom/find-and-replace/pull/769/commits/84ca5600c678e883c9b103f182a597ff179f7b1d

lib/project/result-view.coffee
sullerandras: Fixed in https://github.com/atom/find-and-replace/pull/769/commits/9fe4836a303924e60dfbacbe6ba2090de1f4c0af

lib/project/result-view.coffee
sullerandras: Fixed in https://github.com/atom/find-and-replace/pull/769/commits/3d7e8cd386732ae4737e1de011367585a4399d12

thebarty:
Plus one for getting this to work

dirk-thomas:
Please see #847 for another working proposal.

dirk-thomas:
With #847 being merged this can probably be closed.

Issue #568, opened 2015-10-22
Introduce "Find All" button


mlalkaka:
(This is really a feature request, not technically a defect.)
As a new user to Atom, I recently spent a really long time trying to figure out how to perform the "Find and Replace: Find All" action in Atom. I found the "Find and Replace: Select All" action, but it didn't work for things I typed into the Find and Replace pane, and I just didn't realize there was a different action called "Find and Replace: Find All". 

It would have really been nice to have a "Find All" button in the Find and Replace pane, similar to Sublime Text.

simurai:
This got recently added part of https://github.com/atom/find-and-replace/pull/509. Although it just says "Find". You can already try it in [Atom Beta](https://atom.io/beta), but will eventually also be in Atom stable.

To do a "Find All" without pressing a button and use the keyboard instead, you can type your search term and then hit <kbd>enter</kbd>. "Replace All" would be <kbd>cmd</kbd><kbd>enter</kbd>.

mlalkaka:
Thanks for looking into this @simurai . However, I couldn't figure out how #509 solves the problem. I tried the Atom Beta, and I could only see a "Find" button. Typing a search term and hitting <kbd>enter</kbd> just performed a "Find Next" operation, not a "Find All" operation. Am I missing something? Thanks.

simurai:
Just to be sure, are you trying to do a "Find All" for the current file <kbd>cmd</kbd><kbd>F</kbd> or all files in the entire project <kbd>cmd</kbd><kbd>shift</kbd><kbd>F</kbd>?

mlalkaka:
I'm trying to do a "Find All" (find-and-replace:find-all) in the current buffer/file. The shortcut <kbd>alt</kbd>+<kbd>enter</kbd> exists, but there's no button, which made it difficult to discover as a new user.

simurai:
> in the current buffer/file.

Oh.. sorry. Now I understand. I didn't even know about <kbd>alt</kbd>+<kbd>enter</kbd> which proves your point. :smile: I'll reopen this issue again.

mlalkaka:
Thanks!

blakebrauer:
I'm interested in implementing this feature but I am unsure of where the find all button would fit in the UI. Currently the option buttons take up the space where I think the find all button would make sense. Any suggestions on the layout?

blakebrauer:
Maybe like this:
[...]

PR #785, opened 2016-09-23
Introduce "Find All" button

blakebrauer:
These changes address issue [#568](https://github.com/atom/find-and-replace/issues/568) by adding a "Find All" button to the file-view find and replace as well as to the Find menu.

38aeac60-3f8f-11e6-9bac-1a96702b9a9c

screenshot from 2016-09-23 13 07 34

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -39,15 +39,15 @@ class FindView extends View
         @div class: 'input-block-item', =>
           @div class: 'btn-group btn-group-find', =>
             @button outlet: 'nextButton', class: 'btn', 'Find'
+          @div class: 'btn-group btn-group-find-all', =>
+            @button outlet: 'findAllButton', class: 'btn', 'Find All'
+
+        @div class: 'input-block-item option-block', =>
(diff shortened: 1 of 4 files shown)


blakebrauer:
Could we look into getting this merged?

maxbrunsfeld:
/cc @simurai @blakebrauer has implemented the UI that you [discussed in #568](https://github.com/atom/find-and-replace/issues/568#issuecomment-230070854):

unwrapped

I wanted to merge this PR because I plan on making major changes to this package in order to remove jQuery. However, I noticed that this PR does introduce a problem when the window gets narrow. The 4 search option buttons, which should probably remain grouped, are now separated:

wrapped

I'm not super familiar with the CSS involved. When you get a chance, could you take a look at the markup and CSS here, and make sure it degrades well in a small window?

simurai:
@maxbrunsfeld In https://github.com/atom/find-and-replace/pull/839 the inputs are kept above each other and all the buttons wrap. So that the option buttons are still next to each other:

screen shot 2017-01-08 at 8 04 04 pm

PR #789, opened 2016-10-07
Stop using shadow DOM selectors

as-cii:
We are in the process of removing the shadow DOM boundary from `atom-text-editor` elements. This pull request upgrades existing selectors so that they:

* Stop using `:host`.
* Stop using `atom-text-editor::shadow`.
* Prepend syntax class names with `syntax--`.

/cc: @simurai

### DIFF
--- a/styles/find-and-replace.less
+++ b/styles/find-and-replace.less
@@ -2,8 +2,7 @@
 @import "syntax-variables";
 
 // result markers
-atom-text-editor,
-atom-text-editor::shadow {
+atom-text-editor {
   .find-result .region {
     background-color: transparent;
     border-radius: @component-border-radius;
@@ -26,8 +25,7 @@ atom-text-editor::shadow {
 }
 
 atom-workspace.find-visible {
-  atom-text-editor,
-  atom-text-editor::shadow {
+  atom-text-editor {
     .find-result,
     .current-result {
       display: block;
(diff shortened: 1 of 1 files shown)


Alhadis:
@as-cii This is raising a warning in Deprecation Cop:

> **In styles/find-and-replace.less:** Style elements within text editors using the `atom-text-editor::shadow` selector or the `.atom-text-editor.less` file extension.
If you want to target overlay elements, target them directly or as descendants of `atom-overlay` elements.

Is this to be expected?

50Wliu:
@Alhadis make sure that if you have fnr apm linked that you've pulled from matter recently.

Alhadis:
Erm. I understood everything except the `fnr apm` part...

50Wliu:
fnr -> find-and-replace.  And I really should have done `apm link` instead to make it clearer that fnr and apm were unrelated.

Alhadis:
Oh! Right. Well I have, which is why I'm confused...

as-cii:
Thanks for the report, @Alhadis! I think you might be using a non-dev version of Atom (1.11 stable or 1.12 beta) and have this package linked locally at the same time. Linking newer versions of bundled packages with older versions of Atom is unsupported and could cause _undefined behavior_. Specifically, I think Atom stable/beta is running some heuristics that made sense when we still used shadow DOM boundaries for `<atom-text-editor>` elements; recently, however, we have removed such boundaries from editor elements, which is what's causing the older version of Atom you're using to get confused. 

If that's the case, I would suggest either building Atom from master or unlinking this package and wait  until it gets bundled in a newer official Atom build (1.13 beta/stable). 👍

PR #726, opened 2016-05-23
Add logic to consume file-icons service

Alhadis:
[Same affair as before](https://github.com/atom/fuzzy-finder/pull/211). Hooks the `find-and-replace` package up to the icon-service, allowing packages to customise the file-icons that're displayed in project search results.

PR #792, opened 2016-10-12
Fix breakage if icon-service is passed non-disposable

Alhadis:
Errors started being thrown after upgrading to Atom 1.11.0, and I noticed this was the culprit.

This PR patches #726 to avoid raising if a service-provider fails to return an `onWillDeactivate` disposable (or lacks an `onWillDeactivate` handler altogether).

**References:** atom/event-kit@d1ef613

### DIFF
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -116,7 +116,7 @@ module.exports =
 
   consumeFileIcons: (service) ->
     FileIcons.setService service
-    @subscriptions.add service.onWillDeactivate ->
+    @fileIconsDisposable = service.onWillDeactivate ->
       FileIcons.resetService()
 
   provideService: ->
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Sorry I didn't think to ask about this earlier: what is the `onWillDeactivate` method for? From the name, it seems like a hook that will be called when the package that provides the `atom.file-icons` service is deactivated. But that functionality is already built into the service system. You can simply return a disposable from your service hook:

```coffee
  consumeFileIcons: (service) ->
    FileIcons.setService(service)
    new Disposable -> FileIcons.resetService()
```

Alhadis:
To be honest, I've only included it to stay consistent with the APIs of the [`tree-view`](https://github.com/atom/tree-view#api) and [`tabs`](https://github.com/atom/tabs#api) packages...

Issue #758, opened 2016-07-19
Copy, Find, Paste pastes previous value of find input


ssorallen:
### Prerequisites

1. Create a new TextEditor with the contents
```
foo
bar
```
1. Open `find-and-replace` <kbd>Ctrl-f</kbd>.
1. In the `find-and-replace` panel search for the text `foo`.
1. Expected and actual `foo` is found.
1. Highlight the word `bar`.
1. Press <kbd>Ctrl-c</kbd> and <kbd>Ctrl-f</kbd> in rapid succession. It helps to hold down the <kbd>Ctrl</kbd> and quickly press c and f.

**Expected:** `find-and-replace` to contain the word `bar` because that is highlighted and `find-and-replace` defaults to the highlighted value.

**Actual:** `find-and-replace` changes to the word `bar` and then back to `foo`.

### Original report
***

A muscle memory interaction I've become used to in other text editors is to select some text, copy it (cmd-c), focus find (cmd-f), and paste (cmd-v). If you do this quickly in Atom, you end up with the previous value that was in the find-and-replace input field.

1. Open find-and-replace (cmd-f)
2. Type "foo"
3. Open a file
4. Highlight some text (let's say "bar" for reproducibility)
5. *Quickly* copy, focus find, and paste (cmd-c, cmd-f, cmd-v)

I expect "bar" to be in the find-and-replace input because that's what I copied and pasted, but I reliably get "foo".

lee-dohm:
I just tried this with Atom v1.9.9 on Mac OS X 10.11.6 and was unable to reproduce the problem. Can you verify that this still happens for you? (It is possible that I'm just not doing it fast enough, I suppose.)

ssorallen:
@lee-dohm Thanks for checking, and yes this is still happening to me on Atom 1.9.9. I am on a dual-core machine, which is maybe constrained enough to reliably reproduce the problem?

cmd-c, cmd-f, cmd-v. I keep cmd pressed the whole time and quickly press c, f, v.

iolsen:
Speed seems to matter. If you're not really fast in the copy-find-paste key sequences, this doesn't happen. I'm able to repro on 1.9.9/10.11.6.

ssorallen:
@iolsen Yeah, you have to be quick. It's the same sequence I have used in other text editors and in Chrome when copying a term and searching for it on the page, which is why this still hits me frequently. I use the same pattern in multiple other applications.

vjeux:
[...]

Issue #121, opened 2014-01-10
Cursor doesn't behave properly in find editor


probablycorey:
1. Press `cmd-f`
2. In the find editor type `()`
3. Place the cursor inside the parens.
4. Type `x`, wait a second and notice that the cursor moved outside the parens.

PR #794, opened 2016-10-14
Fix cmd-c cmd-f quick sequence from altering the text

vjeux:
There is currently a very annoying bug in atom where if you are quick to search, it will override the search with the previous value. See https://github.com/atom/find-and-replace/issues/758 for the bug report and repro steps.

Looking at the blame, it looks like the behavior to replace the text was there since the very early days but in 2014, a check was added to make sure that it doesn't override it if the text is the same as it would make the cursor jump ( https://github.com/atom/find-and-replace/issues/121 ).

I don't understand why when the result of a search comes back, you want to modify the input text. I do think this is a bug and the behavior should be to just leave the text as is and when that new text is finished searching it'll update the search results. So, this removes the entire update logic.

Test Plan:
Do the repro steps of https://github.com/atom/find-and-replace/issues/758

Before this commit:

After this commit:

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -313,9 +313,6 @@ class FindView extends View
     else
       @clearMessage()
 
-    if @model.getFindOptions().findPattern isnt @findEditor.getText()
-      @findEditor.setText(@model.getFindOptions().findPattern)
-
   findError: (error) =>
     @setErrorMessage(error.message)
 
@@ -426,7 +423,9 @@ class FindView extends View
     if editor?.getSelectedText?
       findPattern = editor.getSelectedText() or editor.getWordUnderCursor()
       findPattern = Util.escapeRegex(findPattern) if @model.getFindOptions().useRegex
-      @search(findPattern) if findPattern
+      if findPattern
+        @findEditor.setText(findPattern)
+        @search()
 
   findNextSelected: =>
     @setSelectionAsFindPattern()


vjeux:
(I'm looking at the test failures)

vjeux:
Explanation on the way I fixed the tests:

Four tests were broken, here's one of them:

```coffee
    describe "when find-and-replace:find-next-selected is triggered", ->
      it "places the selected text into the find editor and finds the next occurrence", ->
        editor.setSelectedBufferRange([[0, 9], [0, 13]])
        atom.commands.dispatch workspaceElement, 'find-and-replace:find-next-selected'

        expect(findView.findEditor.getText()).toBe 'sort'
        expect(editor.getSelectedBufferRange()).toEqual [[1, 6], [1, 10]]
```

What happened is that `find-and-replace:find-next-selected` command used to do a search and once it completed, update the find input. Because we removed that code path, it doesn't do it anymore.

Looking at the implementation of `find-and-replace:find-next-selected`, it calls a method called `setSelectionAsFindPattern` which actually just triggered a search and not did what it was named to do.

The fix is to make `setSelectionAsFindPattern` actually set the selection as the find pattern and trigger a search. I believe this is the right thing to do as before it would kind of do it if no race conditions appeared.

maxbrunsfeld:
@vjeux Thanks so much for looking into this, and for the great explanation. I just did some manual testing and confirmed that this fixes the issue. Also, I agree that no additional automated tests are needed for this, since the fix was to remove some code that was completely wrong.

vjeux:
@maxbrunsfeld yay, thanks a lot! My first contribution to atom <3

For what its worth, I did try to write a regression test but I couldn't manage to reproduce the behavior in the testing environment. I haven't found a way to wait for the search results to come back in and assert that it doesn't have the wrong value anymore. Here's my attempt for posterity :)

```coffee
    fit "doesn't have race conditions", ->
      editor.setSelectedBufferRange([[2, 8], [2, 13]])
      atom.commands.dispatch editorView, 'find-and-replace:show'

      waitsForPromise ->
        activationPromise

      runs ->
        findView.model.onDidUpdate ->
          console.log 'onDidUpdate'
#          expect(findView.findEditor.getText()).toBe('sort')
          console.log 'getText' + findView.findEditor.getText()

        expect(findView.findEditor.getText()).toBe('items')

        editor.setSelectedBufferRange([[1, 6], [1, 10]])

        atom.commands.dispatch editorView, 'core:copy'
        atom.commands.dispatch editorView, 'find-and-replace:show'
        expect(findView.findEditor.getText()).toBe('sort')
        setTimeout.calls.forEach (c) ->
          console.log c.args[1]
          if c.args[1] is 300
            c.args[0]()
```

cyaris:
what file should the code be added to? thanks.

PR #796, opened 2016-10-15
Don't show find panel when doing cmd-e

vjeux:
There's a long standing issue where cmd-e shows the find panel where it shouldn't: https://github.com/atom/find-and-replace/issues/332

This pull request prevents the find panel from showing.

Test Plan:

1) Click on `new`
2) Press `cmd-e`, note that there is no visual feedback
3) Press `cmd-g` multiple times to go through occurrences

See the behavior in Sublime:

In Atom with this pull request:

It even works the same way in Chrome!


maxbrunsfeld:
✨  Great. Thanks @vjeux.

vjeux:
@maxbrunsfeld <3

maxbrunsfeld:
This will be out in the next beta.

Issue #761, opened 2016-07-23
Add CSS class based on results


simurai:
This came up in https://github.com/atom/one-dark-ui/issues/150.

## Problem

Currently themes have no way to style the find-and-replace panel based on if results are found or not.

## Possible solution

Add a CSS class. E.g:

- `has-results`
- `has-no-results`

Maybe on the root `find-and-replace` element.

![find](https://cloud.githubusercontent.com/assets/378023/16872521/3a3ddce2-4ac8-11e6-880e-3704026aed3c.gif)

PR #798, opened 2016-10-21
Add css class based on if results were returned from a find

zoo1:
enhancement for #761

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -307,6 +307,7 @@ class FindView extends View
     if @model.getFindOptions().findPattern
       results = @markers.length
       resultsStr = if results then _.pluralize(results, 'result') else 'No results'
+      @.addClass(if results then 'has-results' else 'has-no-results')
       @setInfoMessage("#{resultsStr} found for '#{@model.getFindOptions().findPattern}'")
       if @findEditor.hasFocus() and results > 0 and atom.config.get('find-and-replace.scrollToResultOnLiveSearch')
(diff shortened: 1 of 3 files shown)


simurai:
Thanks for improving this. 👍 

We probably should add some specs to make sure the classes don't get removed. I don't know too much about adding specs.. but does adding

```coffee
describe "when no results are found", ->
  it "adds a .has-no-results class", ->
    findView.findEditor.setText 'notinthefile'
    atom.commands.dispatch(findView.findEditor.element, 'find-and-replace:confirm')
    expect(findView).toHaveClass 'has-no-results'

describe "when results are found", ->
  it "adds a .has-results class", ->
    findView.findEditor.setText 'items'
    atom.commands.dispatch(findView.findEditor.element, 'find-and-replace:confirm')
    expect(findView).toHaveClass 'has-results'
```

maybe [here](https://github.com/atom/find-and-replace/blob/master/spec/find-view-spec.coffee#L351) make sense?

simurai:
Another thing, the classes should get removed when either condition is not met.

search

Currently the `has-no-results` and `has-results` stay until Atom gets reloaded.

lib/find-view.coffee
```diff
@@ -307,6 +307,7 @@ class FindView extends View
     if @model.getFindOptions().findPattern
       results = @markers.length
       resultsStr = if results then _.pluralize(results, 'result') else 'No results'
+      @.removeClass('has-results has-no-results')
```
50Wliu: Use `this` instead of `@` when it would otherwise be standalone.

simurai:
Ok, this seems ready to 🚢 

Thanks @zoo1

PR #799, opened 2016-10-22
Add 'Copy Path' to search results context menu

### LINKED ISSUE #783: Ability to copy path of search result
When using the search in directory feature, all the results are shown in collapsible sections, but there's no way to copy the path to the file unless you go to 'reveal in tree' and then right click the file and copy the path.

Can the collapsible tabs have 'Copy Path' added to the split/close pane options?

Edit: or perhaps just allow CTRL+C to copy the path when the collapsed section is highlighted.

ssonal:
Closes #783


maxbrunsfeld:
Thanks @ssonal. This will go out in Atom 1.14.

ssonal:
Cheers!

Issue #803, opened 2016-10-29
slow UI when entering a common regex


dansgithubuser:
Enter the following regex in a 100KB file: cout|
It's not a typical intended search, but it is a common unintended search because results are highlighted as you create your regex.
I find it similar to this issue, perhaps insights can be had: https://github.com/atom/fuzzy-finder/issues/156

PR #805, opened 2016-11-01
don't do live search if regex matches empty string

dansgithubuser:
Addresses https://github.com/atom/find-and-replace/issues/803

### DIFF
--- a/lib/buffer-search.coffee
+++ b/lib/buffer-search.coffee
@@ -1,6 +1,7 @@
 _ = require 'underscore-plus'
 {Point, Range, Emitter, CompositeDisposable, TextBuffer} = require 'atom'
 {Patch} = TextBuffer
+FindOptions = require './find-options'
 escapeHelper = require './escape-helper'
 
 ResultsMarkerLayersByEditor = new WeakMap
@@ -57,6 +58,17 @@ class BufferSearch
       layer = editor.addMarkerLayer({maintainHistory: false})
       ResultsMarkerLayersByEditor.set(editor, layer)
     layer
+
+  checkSanity: (findPattern) ->
+    findOptions = new FindOptions(@findOptions.serialize())
+    findOptions.set({findPattern})
+    if regex = findOptions.getFindPatternRegex()
+      try
+        buffer = new TextBuffer('')
+        sane = true
+        buffer.scan regex, ({range}) -> sane = false
+        return sane
+    false
 
   search: (findPattern, otherOptions) ->
     options = {findPattern}
--- a/lib/buffer-search.coffee
+++ b/lib/buffer-search.coffee
@@ -59,16 +59,13 @@ class BufferSearch
       ResultsMarkerLayersByEditor.set(editor, layer)
     layer
 
-  checkSanity: (findPattern) ->
+  doesMatchTheEmptyString: (findPattern) ->
     findOptions = new FindOptions(@findOptions.serialize())
     findOptions.set({findPattern})
     if regex = findOptions.getFindPatternRegex()
-      try
-        buffer = new TextBuffer('')
-        sane = true
-        buffer.scan regex, ({range}) -> sane = false
-        return sane
-    false
+      not regex.test('')
+    else
+      false
 
   search: (findPattern, otherOptions) ->
     options = {findPattern}
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -234,7 +234,7 @@ class FindView extends View
 
   liveSearch: ->
     findPattern = @findEditor.getText()
-    if findPattern.length is 0 or findPattern.length >= atom.config.get('find-and-replace.liveSearchMinimumCharacters') and @model.checkSanity(findPattern)
+    if findPattern.length is 0 or findPattern.length >= atom.config.get('find-and-replace.liveSearchMinimumCharacters') and @model.doesMatchTheEmptyString(findPattern)
       @model.search(findPattern)
 
   search: (findPattern, options) ->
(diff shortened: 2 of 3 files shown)


lib/buffer-search.coffee
```diff
@@ -58,6 +59,17 @@ class BufferSearch
...
+  checkSanity: (findPattern) ->
+    findOptions = new FindOptions(@findOptions.serialize())
+    findOptions.set({findPattern})
+    if regex = findOptions.getFindPatternRegex()
+      try
+        buffer = new TextBuffer('')
+        sane = true
+        buffer.scan regex, ({range}) -> sane = false
+        return sane
+    false
```
maxbrunsfeld: I think this could be simplified to

```coffee
if regex = findOptions.getFindPatternRegex()
  not regex.test('')
else
  false
```

since the `TextBuffer.scan` method just uses the normal JS regex APIs.

lib/buffer-search.coffee
```diff
@@ -58,6 +59,17 @@ class BufferSearch
       ResultsMarkerLayersByEditor.set(editor, layer)
     layer
 
+  checkSanity: (findPattern) ->
```
maxbrunsfeld: I'd prefer this method be named a bit more precisely. How about `doesMatchTheEmptyString(findPattern)`?

maxbrunsfeld:
Thanks for fixing this, @dansgithubuser! The test looks good. I left a couple of minor comments, but otherwise I'd love to merge this.

lib/buffer-search.coffee
dansgithubuser: I picked the name to allow for additional checks to be added to it as needed, but if you think that's unlikely I'm happy to rename.

lib/buffer-search.coffee
dansgithubuser: OK -- I was conservative and tried to match exactly how this is done elsewhere, because I'm not so familiar with the entire project and what's likely to change.

lib/buffer-search.coffee
maxbrunsfeld: Yeah, that's a valid approach. We tend to name things such so that they describe the *current* behavior with good specificity, and then rename them later, if and when their behavior changes.

maxbrunsfeld:
Great! Thanks @dansgithubuser.

50Wliu:
Hey @dansgithubuser.  Currently trying to track down a regression caused by this change (#809).

Is there any reason you had to instantiate a new `FindOptions` object in `doesMatchTheEmptyString` instead of using the existing `getFindPatternRegex` method available in `buffer-search.coffee`?

dansgithubuser:
I've been looking at it for a little while now. I think it's coming back to me. The problem is that doesMatchTheEmptyString needs to take and use a parameter. BufferSearch is designed to search on a hook that triggers when its findOptions are changed. Said another way, you can't ask BufferSearch a question without modifying the state in an undesired manner. So I do what BufferSearch does internally, which is a call to getFindPatternRegex on a FindOptions object.

I should point out, now that I've noticed, that doesMatchTheEmptyString's name should be negated -- doesntMatchTheEmptyString. It was originally checkSanity -- matching empty string is not sane.

dansgithubuser:
I took a look at #809... I'm not sure why \(.*) would be considered a valid regex? I banged it into a random online javascript regex tester: https://regex101.com which also rejected it.

dansgithubuser:
So maybe something like this?

From 96774431c0390e6df2b812d3bcb76e57d70f5acc Mon Sep 17 00:00:00 2001
From: Dan <dansonlinepresence@gmail.com>
Date: Wed, 23 Nov 2016 00:17:31 -0500
Subject: [PATCH] catch exception, doesMatchTheEmptyString ->
 doesntMatchTheEmptyString

---
 lib/buffer-search.coffee | 12 +++++++-----
 lib/find-view.coffee     |  2 +-
 2 files changed, 8 insertions(+), 6 deletions(-)

diff --git a/lib/buffer-search.coffee b/lib/buffer-search.coffee
index 8290f0e..940434c 100644
--- a/lib/buffer-search.coffee
+++ b/lib/buffer-search.coffee
@@ -59,13 +59,15 @@ class BufferSearch
       ResultsMarkerLayersByEditor.set(editor, layer)
     layer
 
-  doesMatchTheEmptyString: (findPattern) ->
+  doesntMatchTheEmptyString: (findPattern) ->
     findOptions = new FindOptions(@findOptions.serialize())
     findOptions.set({findPattern})
-    if regex = findOptions.getFindPatternRegex()
-      not regex.test('')
-    else
-      false
+    try
+      if regex = findOptions.getFindPatternRegex()
+        return not regex.test('')
+    catch e
+      @emitter.emit 'did-error', e
+    false
 
   search: (findPattern, otherOptions) ->
     options = {findPattern}
diff --git a/lib/find-view.coffee b/lib/find-view.coffee
index 8d474a4..8358d62 100644
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -234,7 +234,7 @@ class FindView extends View
 
   liveSearch: ->
     findPattern = @findEditor.getText()
-    if findPattern.length is 0 or findPattern.length >= atom.config.get('find-and-replace.liveSearchMinimumCharac
+    if findPattern.length is 0 or findPattern.length >= atom.config.get('find-and-replace.liveSearchMinimumCharac
       @model.search(findPattern)
 
   search: (findPattern, options) ->
-- 
1.9.1

50Wliu:
Your last diff got cut off :wink:

That looks better though.

dansgithubuser:
Do you want me to make a further PR, or are you good with the cut-off patch?

PR #822, opened 2016-11-30
Remap/remove keybindings to avoid i8n key conflicts

damieng:
As part of our efforts to support international keyboards we are remapping various ctrl-alt keybindings. This change makes the following adjusments to find and replace keybindings on Windows and Linux only;

- `show-replace` keybinding is removed
- `toggle-case-option` changes to <kbd>ctrl</kbd><kbd>shift</kbd><kbd>c</kbd>
- `toggle-selection-option` keybinding is removed
- `toggle-whole-word-option` keybinding is removed

### DIFF
--- a/keymaps/find-and-replace.cson
+++ b/keymaps/find-and-replace.cson
@@ -44,8 +44,6 @@
   'alt-enter': 'find-and-replace:find-all'
   'ctrl-alt-/': 'find-and-replace:toggle-regex-option'
   'ctrl-shift-c': 'find-and-replace:toggle-case-option'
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
Should we just *remove* some of these bindings, and let people bind them themselves?

damieng:
I'm always up for less default keybindings hence the removal of show-replace (you can't use replace without find anyway).

I like the idea of just having case-sensitive.  Do you want to keep all the mac bindings?

maxbrunsfeld:
I think we should keep the mac ones.

PR #829, opened 2016-12-14
Results pane direction option (#585, #597)

HebaruSan:
It would be great to be able to open project-find results in a downward pane as per #585, so I'm attempting to update and address the comments on #597 by @bencolon. Changes from that PR:

- Resolved the conflict in one of the spec files
- Now using a none/down/right schema to match atom/markdown-preview#368 as per review comment by @acontreras89 (allows for future support of left/up if desired)
- A fix for a problem with single-click pending mode that the specs were catching


maxbrunsfeld:
Thanks @HebaruSan and @bencolon! This looks great. It will go out in Atom 1.14.

maxbrunsfeld:
FYI, I have tweaked the name of the setting. It is now `projectSearchResultsPaneSplitDirection`.

PR #836, opened 2017-01-03
Avoid clearing destroyed marker layers when editors are destroyed

maxbrunsfeld:
Refs https://github.com/atom/atom/pull/13523


maxbrunsfeld:
Superseded by https://github.com/atom/find-and-replace/pull/837.

PR #838, opened 2017-01-06
Remove the use of jQuery

maxbrunsfeld:
* [x] Stop using jQuery explicitly in specs
* [x] Convert specs to JS
* [x] Convert specs to use `async`/`await` instead of `waitsForPromise`/`runs`
* [x] Convert implementation to JS
* [x] Convert views to perform initial render w/ etch instead of space pen
* [x] Use CSS animation for wrap icon
* [x] Use editor font when displaying matches

Refs https://github.com/atom/atom/issues/13254


maxbrunsfeld:
@ungb - Could you give this PR a test when you get a chance? A lot of the UI was rewritten.

Things that might be worth testing:
* Selecting search results using the arrow keys or mouse. Does the view scroll correctly?
* Scrolling through the search results using `Page Down` or `Page Up`. If your keyboard doesn't have these keys, you can bind them by putting something like this in your `keymap.cson`:

  ```
  'atom-workspace':
    'cmd-alt-shift-down': 'core:page-down'
    'cmd-alt-shift-up': 'core:page-up'
  ```

* Responsiveness when there are lots of search results. It should be as good or better than it was before.

Ben3eeE:
@maxbrunsfeld I did some testing today and found a few things, added steps to reproduce below together with a gif of reproduction. Let me know if something is unclear and you need more information or if you want me to do some more testing. 

---- 

I tested using Windows 7 64bit, Atom `1.15.0-dev-1a039df x64` and `Atom 1.14.0-beta4 ia32` 
I linked this package using `apm link --dev` started Atom from `cmd.exe` in my local Atom repository using `atom . --dev` 
Before testing I disabled all community packages and cleared my `init.coffee` file. 

---- 

### Extremely slow search leading to crash 

When testing I found one issue where the search was extremely slow which eventually lead to Atom crashing. I found repro steps and recorded a gif. 

1. Open Atom in your local `atom` repostitory 
1. Open `project find and replace` using `Ctrl-Shift-F` 
1. Do a search for `p` 
1. Expected and Actual: Lightning fast search with many results in many files 
1. Close the `Project Find Results` tab 
1. Do a search for `p` 
1. Expected: Lightning fast search with many results in many files, or just open previous search result. Actual: Extremely slow search with steadily increasing memory usage and eventually Atom crashes. To reproduce crash I sometimes had to search again using these steps. 

**Note:** Same issue with other searches. For example if first search is `stuff` and second search is `a`. 

**Gif of Reproduction** 
far search slow 

## No results of undefined 

1. Open Atom in your local `atom` repostitory 
1. Open `project find and replace` using `Ctrl-Shift-F` 
1. Do a search for `p` 
1. Expected and Actual: Lightning fast search with many results in many files 
1. Do a search for nothing by erasing the `p` and pressing `Enter` 
1. Expected: No results found. Actual: No results found for `undefined` 
1. Close the `Project Find Results` tab 
1. Do a search for `p` 
1. Expected: Lightning fast search with many results in many files. Actual: No results found for `p` 

**Gif of Reproduction** 
far results undefined 

## Another case of no results of undefined 

Clicking the `Use Regex`, `Match Case` or `Whole Word` button causes the `Project Find Results` tab to display `No results for 'undefined'` 

1. Open Atom in your local `atom` repostitory 
1. Open `project find and replace` using `Ctrl-Shift-F` 
1. Do a search for `stuff` 
1. Click any of the `Use Regex`, `Match Case` or `Whole Word` buttons 
1. Expected and Actual: Search restarts with the new settings 
1. Change your mind and click the button again 
1. Expected: Search restarts with the new settings Actual: Shows `No results for 'undefined'` until search finishes 

**Gif of Reproduction** 
far no results undefined 

## Focus is different 
When clicking or doubleclicking on results on the `Project Find Results` tab without this branch the corresponding `TextEditor` is always focused and opened. 
With this branch the `TextEditor` is focused only when clicking also opens the file in a pending tab. 

1. Open Atom in your local `atom` repostitory 
1. Close the `windows.md` tab if open 
1. Open `project find and replace` using `Ctrl-Shift-F` 
1. Do a search for `stuff` 
1. Doubleclick on the result in `windows.md` 
1. Expected: `windows.md` is opened and focused. Actual: `Project Find Results` has focus 
1. Close the `windows.md` tab 
1. Click on the result in `windows.md` 
1. Expected and Actual: `windows.md` is opened and focused 

**Gif of Reproduction** 
far focus

lib/project/result-view.js
```diff
@@ -111,47 +111,33 @@ class ResultView {
...
-        )
+        etch.dom(MatchView, {
+          match,
+          key: i,
+          regex: this.regex,
+          replacePattern: this.replacePattern,
+          isSelected: (i === this.selectedMatchIndex),
+          previewStyle: this.previewStyle,
+          top: itemTopPosition
+        })
```
maxbrunsfeld: Curious about your reasoning for this. It seems like now, components like `ResultView` and `MatchView` are much more coupled to the fact that we use them inside of a `ListView`, and the `ListView` has a less generic interface. Was the `div` causing a problem for styling or something?

lib/find-view.js
```diff
@@ -0,0 +1,737 @@
...
+    this.createWrapIcon();
+  }
+
+  update(props) {}
+
+  render() {
+    return (
+      etch.dom.div({tabIndex: -1, className: 'find-and-replace'},
+        etch.dom.header({className: 'header'},
+          etch.dom.span({ref: 'descriptionLabel', className: 'header-item description'},
```
Alhadis: Currently breaking here:

~~~
Uncaught TypeError: etch.dom.span is not a function @ find-view.js:33
	render @ find-view.js:33
	initialize @ component-helpers.js:60
	FindView @ find-view.js:16
	createViews @ find.coffee [sm]:141
	(anonymous function) @ find.coffee [sm]:42
	module.exports.CommandRegistry.handleCommandEvent @ command-registry.coffee:255
	(anonymous function) @ command-registry.coffee:1
	module.exports.KeymapManager.dispatchCommandEvent @ keymap-manager.js:599
	module.exports.KeymapManager.handleKeyboardEvent @ keymap-manager.js:390
	module.exports.WindowEventHandler.handleDocumentKeyEvent @ window-event-handler.coffee:76
	(anonymous function) @ window-event-handler.coffee:1
~~~

maxbrunsfeld: Yeah, thanks for giving this a spin @Alhadis - it depends on a new etch API introduced on this branch: https://github.com/atom/etch/tree/custom-virtual-dom

Alhadis: Oh, sorry. Should I have symlinked that branch to a local copy of the `etch` repo?

maxbrunsfeld: We realized that we needed to make some changes to this branch after @Ben3eeE's [comment](https://github.com/atom/find-and-replace/pull/838#issuecomment-277758842). It's almost ready to go again, but we still need to publish a new version of etch and fix up some tests. We'll probably be done with that soon, but if you're interested in testing before that, yeah, you could link etch and checkout that branch. Thanks!

maxbrunsfeld:
@Ben3eeE @ungb I believe this is once again ready to test.

@Ben3eeE I think we've addressed the great points you made after your first round of testing:
* Memory usage of project search results view
* Incorrect message after clearing/switching your search
* Incorrect focus when clicking on results

Ben3eeE:
@maxbrunsfeld I can confirm that all things I found earlier and more are addressed. The search is ⚡️ fast ⚡️  💯! I found some minor things only:

**Issues introduced by this PR:**

* If you select a result in some file and press `Collapse All` the selected element does not move to the file and stays on the search result. On `master` the selected element moves to the file when you collapse with a search result in that file selected. So navigating with the keyboard after collapse is not working correctly.

* Using `atom-material-ui` the files in the `Project Find Results` are not highlighted for me. They are highlighted on `master` with `atom-material-ui`.

**Also an issue on master:**

* If you do a search for `p` -> a search for nothing -> close the `Project Find Results` -> a search for `p` the `Project Find Results` displays `No results found for 'p'` instead of searching for `p`.

* If you edit a file during an ongoing search the `Collapse/Expand All` buttons become visible.

maxbrunsfeld:
@Ben3eeE Thanks for catching that regression with `atom-material-ui`. I opened a PR on that theme to use a more flexible selector in one place. There may be more issues like this, but I don't think we can avoid changing the markup in this case, so we're just going to have to be ready to help people when breakage occurs.

Alhadis:
Shit. I better run File-Icons' tests again...

Ben3eeE:
@maxbrunsfeld 

**Issue select a result and `Collapse All`**

* Now if you select a result in a file and `Collapse All` the highlight moves to the file but I am unable to expand it with right arrow key unless I first move to another file.

**New issues found**

* Collapse/Expand state is remembered between searches. So if you collapse the first file and search for something else the first file is collapsed in the next search result. Even if that is for a different file.
* In my dev tools console I see the following warning: `
C:\Users\lineri\Desktop\issues\atom master\find-and-replace\lib\project\results-view.js:33 IntersectionObserver.observe(target): target element is not a descendant of root.`. This happens when I close the `Project Find Results` before performing a new search.

ungb:
Thanks @Ben3eeE for all the thorough testing!  I went through some ad-hoc testing and tried some of the scenarios you mentioned above and didn't see any issues.  LGTM!

Ben3eeE:
@maxbrunsfeld @as-cii  Really great work :100: I can't find any other issues that were introduced by this PR :ship:

PR #839, opened 2017-01-07
[WIP] Fix wrapping on narrow windows

simurai:
### Description of the Change

Fixes the wrapping when the window gets narrow.

Before | After
--- | ---
wrapped | screen shot 2017-01-08 at 8 04 04 pm

### Benefits

Keeps the options button grouped together on narrow windows.

### Possible Drawbacks

It's now a bit different than the project search. Also markup/css don't share as much anymore.

### Applicable Issues

This is a follow up to: https://github.com/atom/find-and-replace/pull/785#issuecomment-270996274

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -40,8 +40,6 @@ class FindView extends View
             @button outlet: 'nextButton', class: 'btn', 'Find'
           @div class: 'btn-group btn-group-find-all', =>
             @button outlet: 'findAllButton', class: 'btn', 'Find All'
-
-        @div class: 'input-block-item option-block', =>
           @div class: 'btn-group btn-toggle btn-group-options', =>
             @button outlet: 'regexOptionButton', class: 'btn', =>
               @raw '<svg class="icon"><use xlink:href="#find-and-replace-icon-regex" /></svg>'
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
> It's now a bit different than the project search.

I didn't think of this downside at first, but I think I'm ok with it, since the project search buttons were already somewhat different (no `Replace` button, just `Replace All`). @simurai Overall, what's your feeling about this; 🚢 ? If not, I can revert #785.

simurai:
Hmm.. I'm torn on this.

Another problem: The "Only in Selection" and "Whole Word" options might make it seem they are part of "replace" because they are in the same row as the replace input and buttons.
 
screen shot 2017-01-10 at 2 36 20 pm

Maybe we could group them in another way. Or have them above the find buttons, smaller next to "Finding with Options: Case Insensitive".

screen shot 2017-01-10 at 3 02 48 pm

I'll see how easy it is to move them.

simurai:
@maxbrunsfeld Ok, the option buttons are moved into the header in https://github.com/atom/find-and-replace/pull/840.

I think we should merge that one and close this one.

maxbrunsfeld:
Superseded by #840.

PR #840, opened 2017-01-10
Move options into the header

simurai:
### Description of the Change

This moves the option buttons into the header.

#### Before

screen shot 2017-01-10 at 9 21 46 pm

#### After

screen shot 2017-01-10 at 9 18 03 pm

Also the "Find in project" options got are moved to stay consistent:

screen shot 2017-01-10 at 9 33 39 pm

Note: The buttons are a bit large, but that's an issue with the One themes.

### Benefits

Keeps the options button grouped together in a row and doesn't give the impression that the bottom two are part of "replace".

### Possible Drawbacks

- The header text starts wrapping earlier
- Initially will mess with people's muscle/visual memory.

### Applicable Issues

This is a follow up to: https://github.com/atom/find-and-replace/pull/785#issuecomment-270996274

And an alternative to: https://github.com/atom/find-and-replace/pull/839

### DIFF
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -28,6 +28,15 @@ class FindView extends View
         @span class: 'header-item options-label pull-right', =>
           @span 'Finding with Options: '
           @span outlet: 'optionsLabel', class: 'options'
+          @span class: 'btn-group btn-toggle btn-group-options', =>
+            @button outlet: 'regexOptionButton', class: 'btn', =>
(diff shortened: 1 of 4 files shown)


maxbrunsfeld:
:+1: Wow I really like how this preserves greater consistency between buffer search and project search. Great idea.

50Wliu:
I like this as well!

maxbrunsfeld:
Thanks for doing this @simurai!

PR #847, opened 2017-01-29
add settings for context lines, show context in match view

dirk-thomas:
### Requirements

* atom/atom#13713
* atom/scandal#41
* atom/text-buffer#195

### Description of the Change

The patch adds two new options `searchContextLineCountBefore` and `searchContextLineCountAfter` which are passed down to `scandal` to return matches with context lines around the matched line. The match view is being modified that each match does not contain a single `li` but a whole list which itself contains all context lines before, the matching line, and the context lines after.

To allow the context lines and the match line to be distinguished visually the color of the match line has been updated. Also the way the ".selected" item is being determined is updated to highlight the match line rather then the first context line in the sub list.

### Alternate Designs

Other approaches:
* #702
* #769

In contrast to these this PR does not rely on requiring a `TextBuffer` for each search result. Instead it uses an extended version of `scadal` to provide the context directly in the returned match objects.

### Benefits

It implements the long desired feature described in #190. The approach requires less resources as the listed other approaches.

### Possible Drawbacks

The new settings default to `0` lines of context before and after which is the same behavior as previously. But since the view structure has been updated to accommodate the optional context lines that might affect compatibility. I don't have any experience with Atom views and therefore can't estimate how high the chances for this are.

### Applicable Issues / Remaining Todos

* ~~In the current state when the search results list a file which is not opened yet and you click on the result the match is being updated and looses the context information. This is due to the fact that the search is being repeated in the buffer (instead of using `scandal` through the workspace API. In order to address this the `TextBuffer.scan` needs to be extended to support the same new options as in the referenced `scandal` ticket. Therefore I have prefixed the title with `[WIP]` for now.~~ This has been address with the PR to the `text-buffer` package.

* The current settings don't use a custom `order` attribute and are therefore in alphabetical order. Since it is a bit confusing that the `before` option comes after the `after` option it might be good to add explicit `order` attributes to all options.

* Add test coverage / update existing specs as soon as there is a consensus on the selected approach and the feature works.

### Related Stuff

The following items would all be nice to have but already a problem in the current state and I consider them orthogonal to this PR. They can be improved / addressed in the future in separate PRs:

* Correct indentation (multiple whitespace in each line are only rendered as a single space)
* Multiple results in a single line (and now also overlapping lines of context) could be deduplicated
* Gaps between code in the same file could be expandable


dirk-thomas:
I removed the "[WIP]" prefix from the title since the feature is "complete" now. While there are still failing specs to address I think it would be good to review this first to ensure there is a consensus on the approach.

maxbrunsfeld:
@dirk-thomas Thanks for this! Could you provide screenshots or animated GIFs showing what this new UI look like?

One concern: I'm not sure that a config setting is the best way for the user to specify how many context lines they want. I would think that you'd want to vary these numbers on a search-by-search basis. For me personally, I usually wouldn't want *any* lines of context, but in certain searches, I'd want to see some preceding and/or following context, depending on what I'm looking for. So I would think that we might actually want some kind of UI for specifying the amount of context.

What do you think?

/cc @simurai @atom/feedback

dirk-thomas:
I kept the UI changes minimal for now so that the PR is easier to review. Please see the current results with 2 lines of context above and below. The matched line is highlighted in white (new to better distinguish the matched lines from the context line) and the matched text is highlighted with a blue box (as it was before).

search-results-with-context

While the PR only adds package settings for the number of context lines the API is designed to provide the arguments for each query. So it should be easy to customize the UI to configure it for every search. I am not familiar with the UI elements in Atom. Maybe someone could suggest how to setup the additional elements beside the existing search options.

For the search results I could see a few enhancements to improvement the view (specific to the new context):
* reduce spacing between the context lines belonging to a single result (maybe same spacing as lines in a TextEditor?)
* more separation between search results (currently only uses a 2px padding - same as before), or a empty line, or a horizontal line for visual separation
* optionally fold / unfold the context (that might another way to address the use case you mentioned, where you only sometimes need to see the context)

Besides there a other possible improvements but they also apply to the result without the context:
* render indentation of lines correctly
* deduplicate results, if multiple matches are found in the same line

50Wliu:
> a horizontal line for visual separation

I like this option the best.  Just scanning the screenshot it's very hard to tell where a jump in the file occurs.

dirk-thomas:
@50Wliu I added a second commit (747c91e91b284d3c941c4fbc72f7ae6dbfe89c1d) which simply modifies the css. The result is definitely easier to "parse" visually:

search-results-with-context -with-sep-line

simurai:
> @maxbrunsfeld I usually wouldn't want any lines of context, but in certain searches, I'd want to see some preceding and/or following context, depending on what I'm looking for.

What about something similar like the "blue expanding button" on github.com. So you could get context for a specific line. Maybe each click increases context by 5 extra lines.

expand

Then there could also be a `<input class="input-number" type="number" />` in the header where you can update the config for all lines. So that you don't have to open the package settings every time. Not sure if it needs a separate value for before and after?

number

dirk-thomas:
@simurai The advantage of this PR over the other approaches is that it relies on `scandal` to return the context lines together with each match. That is more efficient because it does not require a text buffer to be opened for every result (consider searching through many large files). But as a consequence you can't just add "more" context since that requires to open the file containing the match again and reading the additional lines.

As you mentioned it could be considered in a follow up PR.

simurai:
> @dirk-thomas requires to open the file containing the match again and reading the additional lines.

Oh, I see. Ok, another naive idea :innocent:.. what about **always** get some extra context and then hide it when initially shown? Maybe keep the context in memory or add to the DOM and hide with CSS? Then when expanding it would be instantly available. Of course, this only makes sense if the extra context doesn't significantly slow down search.

mherodev:
@dirk-thomas The horizontal rules are a little noisy visually.

How about using ellipses in place of a line number instead?

image

dirk-thomas:
@simurai The optional folding / unfolding of the context was what I tried to suggest earlier.

@mherodev The ellipses will be trivial to implement instead of the horizontal line. I will rather wait until we have a consensus in which direction this should go before updating the PR. Your picture shows well how it could look like.

dirk-thomas:
@maxbrunsfeld Can you please change the label from "work-in-progress" to "needs-review".

bmherold:
Getting very close to feature parity with Sublime!

The only suggestion I have is to collapse overlapping results into a single block (Sublime does this) which will reduce noise. That may require a separate PR but it's worth mentioning. I attached a screenshot of Sublime for reference.

screen shot 2017-02-02 at 11 43 39 am

dirk-thomas:
Since the beta of 1.15 was just released it would great if this long awaited feature could be integrated (after a decision on how it should "look" like has been made) in order to become available in the next beta (1.16).

lee-dohm:
@dirk-thomas I say go with the horizontal lines and we can iterate on it once it has been merged.

dirk-thomas:
@lee-dohm Sounds good to me. That matches the current state of the patch.

lee-dohm:
@dirk-thomas Everything is good for final review then?

dirk-thomas:
@lee-dohm I would say yes. If this gets a +1 I would update the specs to match the new html structure if that is ok.

lee-dohm:
@dirk-thomas you have a go 🚀

dirk-thomas:
I quick update about the progress: it took me quite a while to understand more of the internals, fix issues of this patch along the way, and update (most of) the specs. I think it is getting pretty close and the only specs still failing are related to the scrolling with `core:page-down` and `core:page-up`.

Before the invariant was that each result had the same height (one line). With the context (sometime there is non if the result is at the top or bottom of a file) this is not true anymore. I changed the implementation of `selectPreviousPage` and `selectNextPage` quite a bit. The existing specs for this behavior (https://github.com/atom/find-and-replace/blob/652ecbba704acf6f2dc9ac20ffb00e78966f0c2f/spec/results-view-spec.coffee#L292-L346) are there not really appropriate anymore. Is it ok to rewrite these two specs which currently rely on exact pixel coordinates to rather check for the appropriate scrolling direction (without implementing the same logic as the tested code)?

Beside that there are a few specs failing with `callback is not a function`. I currently can't reproduce them locally. Maybe I just have sync my local work with the latest master branches again?

lee-dohm:
> Is it ok to rewrite these two specs which currently rely on exact pixel coordinates to rather check for the appropriate scrolling direction (without implementing the same logic as the tested code)?

According to the description of the specs, when executing `core:page-down/up`, the first result on the next/previous page should be at the top. If the "top of the result" is considered to be the first line of context, could you still make the specs rely on pixel positions? I wouldn't want context to be randomly cut off depending on user's settings.

> Maybe I just have sync my local work with the latest master branches again?

Might be a good idea to see if they go away.

dirk-thomas:
> > Is it ok to rewrite these two specs which currently rely on exact pixel coordinates to rather check for the appropriate scrolling direction (without implementing the same logic as the tested code)?

> According to the description of the specs, when executing core:page-down/up, the first result on the next/previous page should be at the top. If the "top of the result" is considered to be the first line of context, could you still make the specs rely on pixel positions? I wouldn't want context to be randomly cut off depending on user's settings.

I updated the spec to remember the position of the first match (using `offset()`) and check that after scrolling a page the newly selected item is on the same position. While the approach is very different it achieves the same goal of ensuring the newly selected item is at the top (when scrolling down).

> > Maybe I just have sync my local work with the latest master branches again?

> Might be a good idea to see if they go away.

Maybe those can't pass until the other PRs are merged and released. Could that be the case? Or is there another way to let the CI jobs use the referenced PRs for testing?

lee-dohm:
Things look good to me. @maxbrunsfeld do you want to give a final 👍 ?

dirk-thomas:
Friendly ping :wink:

maxbrunsfeld:
Hey @dirk-thomas, this looks good, but I've changed some things on `find-and-replace` master. Are you ok with me pushing to your branch to resolve the conflicts?

dirk-thomas:
@maxbrunsfeld Sure, please go ahead. Let me know if I can help with anything.

caesarsol:
I was missing the feature a lot, just found that you were working on this a few hours ago, can't wait to see it merged in the next few days 😉 

Keep up the good work!

maxbrunsfeld:
@dirk-thomas I have started work on merging the changes from master into this branch. The result is [here](https://github.com/atom/find-and-replace/tree/match_with_context).

The merge is not easy because we had to remove the use of jQuery from this entire package. In the process, I converted the source files to JavaScript and (more importantly) switched to a virtual-DOM based rendering strategy, which has significantly improved the performance of scrolling a long list of results.

As for the status of the merge: on the branch I linked above, I have adapted all of your implementation changes to work within the new rendering strategy. The UI pretty much works, but I have not yet incorporated your changes to the (coffee-script) tests into the new (javascript) tests. Do you think you could port over your changes to the new test files?

Sorry for colliding with you here... the removal of jQuery was pressing because it is a prerequisite to [this](https://github.com/atom/atom/pull/13916) ongoing startup performance work.

Let me know if you have more questions!

maxbrunsfeld:
@dirk-thomas Another thing we need to address is potential redundancy between adjacent matches' context lines. For example, if a file contains this text:

```
1| the grey cat
2| scratched the brown dog.
3| cats are mean.
```

And I search for `cat` with 2 lines of context, my search results should look like this:

**1**| the grey **cat**
2| scratched the brown dog
**3**| **cat**s are mean.

Not this:

**1**| the grey **cat**
2| scratched the brown dog
2| scratched the brown dog
**3**| **cat**s are mean.

I'm not sure at what level we should address the redundancy. Maybe it should actually be handled in `scandal` and `text-buffer`. Or maybe it would make sense to do it in the UI. Thoughts?

dirk-thomas:
Regarding the redundancy: please see the example in my original post under "Related Stuff". The problem already exists without this PR when multiple matches are found in the same line, e.g. searching for "the" the same line might appear multiple times:

**1**| **the** quick brown fox jumps over the lazy dog
**1**| the quick brown fox jumps over **the** lazy dog

Thinking about the following possible UX in the future: first only show the lines containing matches, then "unfold" the context on demand, try to deduplicate where possible. I think this is exclusively a UI problem and should therefore be handled in the view. Neither atom/scandal#41 nor atom/text-buffer#195 should need to know about all of this.

Therefore I would suggest to get this improvement merged as-is (since it is a strict improvement and the shortcomings have been there before). I am happy to work on follow-up PRs which improve the "related"stuff" (which should be limited to changed in this package).

--

I didn't want to bring up the potential problem related to deduplicating lines but since you mentioned the topic I want to briefly dump my thoughts here. In the above example with two matches in the same line it would be "straight forward" to only show the line once and depending on where you click jump to the position where the match starts.

But if you consider potentially overlapping matches it becomes ambiguous which match you are referring to. E.g. search in the following line for "OneOne":

**1**| OneOneOneOne

Clearly the search matches three times:

**1**| **OneOne**OneOne
**1**| One**OneOne**One
**1**| OneOne**OneOne**

In separate lines it is trivial to jump to the correct match. For a merged line it becomes ambiguous. You could argue that you jump to the "right" line anyway. But in a perfect world you expect that e.g. the match you clicked on will be selected when you jump to it.

I just wanted to sketch why I think the deduplication might be quite complicated for those corner cases.

dirk-thomas:
I tried updating the specs but notices that the behavior of `pageUp` / `pageDown` has changed. With my PR when scrolling a page down the selected search result was always at the top of the view. Not when you e.g. search for "describe" in this package and trigger `pageDown` multiple times the selected result is not at the top anymore. After a few pages a "random" result in the middle of the view is selected. It is pretty difficult where this behavior broke since the diff is basically unreadable with all the other changes. Maybe you have an idea what you changed during the refactoring?

Also the horizontal line separating results is gone now on your new branch. I am not sure if that was intentional or if that is a regression too?

maxbrunsfeld:
> But if you consider potentially overlapping matches it becomes ambiguous which match you are referring to.

We don't allow overlapping matches so I don't think this will be an issue for us.

> the behavior of `pageUp` / `pageDown` has changed. With my PR when scrolling a page down the selected search result was always at the top of the view.

The behavior should be that the selected result stays in the *same* position within the viewport, not necessarily at the top of the view. This is what happens on master, and I thought that I saw this working fine on my branch after the merge. Let me take another 👀 .

> Also the horizontal line separating results is gone now on your new branch. I am not sure if that was intentional or if that is a regression too?

Yeah, I'd actually rather not introduce the horizontal line yet. The reason is that I'm ambivalent as to whether it's visually helpful, and it complicates the implementation a little bit, because in order to get good scrolling performance, we now are responsible for calculating heights and positions of search results manually. Sorry for not mentioning that.

dirk-thomas:
> > But if you consider potentially overlapping matches it becomes ambiguous which match you are referring to.

> AFAIK, it's built into the search mechanism that we will never produce overlapping matches, so I don't think this will be an issue for us.

So it might become a problem in the future. Once the search mechanism is "fixed" :wink: But we can think about that when that is the case...

> > Also the horizontal line separating results is gone now on your new branch. I am not sure if that was intentional or if that is a regression too?

> Yeah, I'd actually rather not introduce the horizontal line yet. The reason is that I'm ambivalent as to whether it's visually helpful, and it complicates the implementation a little bit, because in order to get good scrolling performance, we now are responsible for calculating heights and positions of search results manually. Sorry for not mentioning that.

That is totally fine by me.

dirk-thomas:
> > the behavior of pageUp / pageDown has changed. With my PR when scrolling a page down the selected search result was always at the top of the view.

> The behavior should be that the selected result stays in the same position within the viewport, not necessarily at the top of the view. This is what happens on master, and I thought that I saw this working fine on my branch after the merge. Let me take another :eyes: .

The current behavior on your branch scrolls by the `clientHeight`. That makes sense imo since the user expects exactly to see "all new content" without "missing any lines" from what they saw before.

But regarding the newly selected match: Let's assume that the search results all have the same height so the same number of context lines, which means no matches close to the top or bottom of a file. And further assume that the client height is not close to a multiple of each results height but something in between, e.g. 3.5 result heights. So after scrolling a page down the newly selected result must be either ~ half result height further up or down than the previously selected item. Therefore currently the selection slowly drifts when triggering pageDown repeatedly. It seems if the fractional part of the client height / result height is < 0.5 it drifts in one direction, if is > 0.5 it drifts in the other direction.

I am not sure what to do about it and what a "better" / expected behavior would be. I just think the current behavior sometimes looks "random" to the user and the specs could only check of the new selection is within a certain range from the previous one which is kind of fuzzy...

maxbrunsfeld:
> I am not sure what to do about it and what a "better" / expected behavior would be. I just think the current behavior sometimes looks "random" to the user and the specs could only check of the new selection is within a certain range from the previous one which is kind of fuzzy...


Yeah, I agree; it's not ideal. But the `page-down` experience was never exactly *great*; until my most recent PR, adding results to the DOM would sometimes cause noticeable lag and inconsistent scrolling. And the issue you're describing here exists on master.

I guess an ideal solution would be to store off some kind of 'goal offset` for the selected match, so that we could continue to round in the correct direction when paging up and down. That said, I don't think we need to do that as part of this PR.

What would you like to do?

dirk-thomas:
The challenge is that the current behavior is not that easy to test in a spec. I could try to update the specs to check for some kind of valid "range" around there the new selection is supposed to be (probably the old location plus / minus the height of the largest possible result height). If you think this is "good enough" for now we can tackle the scrolling behavior separately if that is already the case on master.

maxbrunsfeld:
I think that when testing `page-up` and `page-down`, we can just set `searchContextLineCountBefore` and `searchContextLineCountAfter` to zero. Would that solve this issue?

dirk-thomas:
Probably, but then we wouldn't really test that the behavior still does something reasonable with context :wink:

dirk-thomas:
I updated the `pageUp` / `pageDown` specs significantly. They check now that after executing the page* command:

* the newly selected result is not the same as before
* the vertical offset (position within the visible area) of the newly selected result is close to the previous one
* the vertical position of the newly selected result if greater (pageDown) / lesser (pageUp)

With this state all specs pass for me locally:

> 277 tests, 854 assertions, 0 failures, 0 skipped

I assume the CI won't pass since the other PRs haven't been merged and released yet?

dirk-thomas:
I updated the patch in 47dd521d99052dd87b5e0a4f8ab5ae27cd676113 to use the same option names as in https://github.com/atom/text-buffer/pull/195#discussion_r104802317.

dirk-thomas:
@maxbrunsfeld Is this still being considered for Atom 1.17 beta?

lee-dohm:
I'm going to bring it up in the team meeting first thing on Monday.

lee-dohm:
@dirk-thomas Thanks for all your hard work and patience on this. I talked to the team about this PR in the team meeting this morning. What we would need for this to be merged and go out with v1.17:

1. Resolve the conflicts that now exist
1. CI to pass - I understand a suggestion was made to make the tests conditionally run based on Atom version

I understand that CI will pass soon enough as the features that this PR depend on make it out to Stable on their own. So we fully intend to integrate this PR, it just won't be with v1.17 without additional work.

Let me know if you have any questions!

dirk-thomas:
@lee-dohm What is the timeline for 1.17? When does this need to be ready in order to be merged in time?

Regarding the compatibility with stable (old API) as well as the upcoming beta (new API): as far as I can see the patch would require conditionals in many many lines / places. Is that really the intention here? I would argue that it makes the code much worse to read / maintain. What is the appropriate condition to check for the Atom version I should use all over the place?

lee-dohm:
The idea is not to place version checks in the code but in the tests. The condition to check would be `atom.getVersion() >= '1.17'`. We're planning on cutting v1.17-beta very, very soon.

maxbrunsfeld:
@dirk-thomas What I had imagined was something like:

```js
if (parseFloat(atom.getVersion()) >= 1.17) {
  describe('when context lines are enabled', () => {

    // tests for the context line behavior

  })
}
```

I realize that we'd also need some conditional like this when we call `editor.scan`, because you made a change to the parameters that method takes. Are there other places where conditionals would be required?

dirk-thomas:
@lee-dohm I don't know what exactly "very, very soon" means but probably only a few days? I will try but I can't promise to find the necessary spare time to work on updating this PR today (tomorrow?). If anybody else would like to rebase the PR and add the conditions please feel free to do so. It would be really sad if this great feature would be left out for yet another cycle...

dirk-thomas:
I rebased the patch and used the Atom version to distinguish the signature of the `TextEditor.scan` function in https://github.com/atom/find-and-replace/pull/847/files#diff-eb24277759b68118e9e07ef629d193e8R250

lee-dohm:
Thanks for all your hard work and perseverance on this @dirk-thomas!

dirk-thomas:
:ship: Awesome. Looking forward for the upcoming beta!

cvogt:
@dirk-thomas thank you so much, you rock!!

PR #852, opened 2017-02-07
Check for the existence of new markers before trying to update markers

### LINKED ISSUE #851: Uncaught TypeError: (var)[Symbol.iterator] is not a function
[Enter steps to reproduce:]

1. Enter `show(` in the search field and select a regex search
2. Open any new editor and type some text (or remove some text)

screenshot 2017-02-07 17 32 16

**Atom**: 1.14.0-beta2 x64
**Electron**: 1.3.13
**OS**: Mac OS X 10.12.2
**Thrown From**: [find-and-replace](https://github.com/atom/find-and-replace) package 0.205.1

### Stack Trace

Uncaught TypeError: (var)[Symbol.iterator] is not a function

```
At /Applications/Atom Beta.app/Contents/Resources/app.asar/node_modules/find-and-replace/lib/buffer-search.js:227

TypeError: (var)[Symbol.iterator] is not a function
    at BufferSearch.bufferStoppedChanging (/app.asar/node_modules/find-and-replace/lib/buffer-search.js:227:93)
    at Function.module.exports.Emitter.simpleDispatch (/app.asar/node_modules/event-kit/lib/emitter.js:25:14)
    at Emitter.module.exports.Emitter.emit (/app.asar/node_modules/event-kit/lib/emitter.js:129:28)
    at /app.asar/node_modules/text-buffer/lib/text-buffer.js:1368:25
```

### Commands

```
  2x -6:31.7.0 editor:move-to-first-character-of-line (input.hidden-input)
     -6:31.5.0 core:save (input.hidden-input)
     -6:28.3.0 core:move-down (input.hidden-input)
  2x -6:28 editor:move-to-first-character-of-line (input.hidden-input)
     -6:27.6.0 core:select-down (input.hidden-input)
     -6:27.2.0 core:backspace (input.hidden-input)
     -6:26.5.0 core:save (input.hidden-input)
     -6:25.9.0 editor:move-to-end-of-screen-line (input.hidden-input)
     -6:25.7.0 editor:move-to-first-character-of-line (input.hidden-input)
     -6:25.2.0 editor:select-to-first-character-of-line (input.hidden-input)
  2x -6:24.9.0 core:backspace (input.hidden-input)
     -6:24.4.0 core:save (input.hidden-input)
     -6:24 editor:newline (input.hidden-input)
     -6:22.6.0 core:save (input.hidden-input)
  2x -6:22.3.0 editor:move-to-first-character-of-line (input.hidden-input)
     -0:02.1.0 core:backspace (input.hidden-input)
```

### Non-Core Packages

```
alignment 1.1.5 
api-docs 0.0.6 
atom-autocomplete-php 0.21.0 
atom-beautify 0.29.17 
atom-typescript 10.1.13 
base64 0.3.0 
blade-snippets 0.2.0 
editorconfig 2.2.0 
elasticsearch-client 0.9.5 
file-icons 2.0.13 
language-blade 0.27.1 
language-docker 1.1.7 
language-dots 0.0.4 
language-nginx 0.6.2 
laravel 0.7.1 
linter 1.11.21 
linter-eslint 8.1.0 
linter-js-yaml 1.2.6 
linter-php 1.3.1 
linter-phpcs 1.5.8 
linter-stylelint 3.7.0 
merge-conflicts 1.4.4 
pigments 0.39.0 
pretty-json 1.6.1 
project-manager 3.3.3 
```

50Wliu:
### Requirements

### Description of the Change

There are two changes included in this PR, conveniently separated into two separate commits.
The first change makes updating markers after a buffer change bail early if no new markers are detected, or if it returns false (signifying an invalid/missing find pattern).  This change does two things: 1) it prevents an iterator error when the iterator tries to iterate over `false`, and 2) it prevents the `did-update` event from being emitted, because if nothing changes, then by definition nothing updated.  This change fixes an issue where the error message would be overwritten with "No results for ..." when backspacing.

The second change is a UX improvement.  Consider the following scenario:
1. Open a file and enter `test`.
2. Search for `test` and notice that the find box turns green (using the One themes).
3. Turn on regex mode and add an opening parenthesis to the find pattern.
At this point, there is contradictory information: a red error message is presented, yet the find box is still green, as if there are results.  This is because when an error message is set, it does not remove the `has-results` class, nor does it add the `has-no-results` class.  Now we do both, and the find box turns red as expected.

Finally, I removed what I think is an incorrect line that nulls out the `findError` function as soon as markers are updated.

### Alternate Designs

None.  This seems like a very clean solution to me.

### Benefits

No more errors when typing text while there's an invalid find pattern, in addition to improved UX when switching over to an invalid find pattern.

### Possible Drawbacks

I'm pretty confident that there will be none.

### Applicable Issues

Fixes #851

### DIFF
--- a/lib/buffer-search.js
+++ b/lib/buffer-search.js
@@ -224,6 +224,8 @@ class BufferSearch {
       }
 
       const newMarkers = this.createMarkers(scanStart, scanEnd);
+      if(newMarkers === false || !newMarkers.length) return; // Invalid/missing find pattern or no new markers
+
       const oldMarkers = this.markers.splice(spliceStart, (spliceEnd - spliceStart) + 1, ...newMarkers);
       for (let oldMarker of oldMarkers) {
         oldMarker.destroy();
--- a/lib/find-view.coffee
+++ b/lib/find-view.coffee
@@ -307,7 +307,6 @@ class FindView extends View
       atom.beep()
 
   markersUpdated: (@markers) =>
-    @findError = null
     @updateResultCounter()
     @updateReplaceEnablement()
 
@@ -342,6 +341,8 @@ class FindView extends View
     @descriptionLabel.text(infoMessage).removeClass('text-error')
 
   setErrorMessage: (errorMessage) ->
+    this.removeClass('has-results')
+    this.addClass('has-no-results')
     @descriptionLabel.text(errorMessage).addClass('text-error')
 
   clearMessage: ->


lib/buffer-search.js
```diff
@@ -224,6 +224,8 @@ class BufferSearch {
       }
 
       const newMarkers = this.createMarkers(scanStart, scanEnd);
+      if(newMarkers === false || !newMarkers.length) return; // Invalid/missing find pattern or no new markers
```
maxbrunsfeld: Even if there are no *new* markers, we still might want to destroy some old markers, and we certainly want to continue processing the other entries in `changes`, because they might cause us to find new markers.

For example, suppose your editor was in this configuration (cursors represented by `|`):

```
1.   one
2.   two|
3.   three
4.   foo|ur
```

and you were searching for the pattern `two|four` and typed a `backspace`. That would *destroy* the match on line 2 but *create* a match on line 4.

In that situation, we wouldn't to return early from `bufferStoppedChanging` just because we didn't find any new markers in the first changed region.

lib/buffer-search.js
```diff
@@ -237,7 +239,7 @@ class BufferSearch {
       }
     }
 
-    this.emitter.emit('did-update', this.markers.slice());
+    if(changes.length) this.emitter.emit('did-update', this.markers.slice());
```
maxbrunsfeld: :+1: This makes a lot of sense.

lib/buffer-search.js
maxbrunsfeld: I think what we might want to do is make the `createMarkers` return an empty array if no matches were found in the given range.

maxbrunsfeld: We've fixed this as part of https://github.com/atom/find-and-replace/pull/860. Didn't mean to steal your 🌩  but we're on a spree of uncaught exception fixes after publishing 1.14.

maxbrunsfeld: I think the other part of this PR is still worthwhile though!

50Wliu: Yeah I thought I noticed a regression related to that, but didn't realize what the steps to reproduce were.

I'll update this PR to only focus on the latter changes.

lib/buffer-search.js
```diff
@@ -237,7 +237,7 @@ class BufferSearch {
       }
     }
 
-    this.emitter.emit('did-update', this.markers.slice());
+    if (changes.length) this.emitter.emit('did-update', this.markers.slice());
```
50Wliu: Should this still be checking for `changes` then, or should it be checking for changed markers? (Not sure how to do that)

PR #884, opened 2017-04-15
Add spec for context lines

dirk-thomas:
### Requirements

* To be filled out when I get this actually working...

### DIFF
--- a/spec/results-model-spec.js
+++ b/spec/results-model-spec.js
@@ -35,6 +35,8 @@ describe("ResultsModel", () => {
       expect(resultsModel.getPathCount()).toBe(1);
       expect(resultsModel.getMatchCount()).toBe(6);
       expect(resultsModel.getPaths()).toEqual([editor.getPath()]);
+      // the following condition is pretty hacky
+      // it doesn't work correctly for e.g. version 1.2
       if (parseFloat(atom.getVersion()) >= 1.17) {
         expect(result.matches[0].leadingContextLines.length).toBe(1);
         expect(result.matches[0].leadingContextLines[0]).toBe("var quicksort = function () {");
@@ -56,6 +58,8 @@ describe("ResultsModel", () => {
       expect(resultsModel.getMatchCount()).toBe(1);
       expect(resultsModel.getPaths()).toEqual([editor.getPath()]);
       expect(result.matches[0].lineText).toBe("there are some items in here");
+      // the following condition is pretty hacky
+      // it doesn't work correctly for e.g. version 1.2
       if (parseFloat(atom.getVersion()) >= 1.17) {
         expect(result.matches[0].leadingContextLines.length).toBe(0);
         expect(result.matches[0].trailingContextLines.length).toBe(0);
--- a/spec/results-model-spec.js
+++ b/spec/results-model-spec.js
@@ -10,6 +10,8 @@ describe("ResultsModel", () => {
 
   beforeEach(async () => {
     atom.config.set("core.excludeVcsIgnoredPaths", false);
+    atom.config.set("find-and-replace.searchContextLineCountBefore", 2);
+    atom.config.set("find-and-replace.searchContextLineCountAfter", 3);
     atom.project.setPaths([path.join(__dirname, "fixtures")]);
 
     editor = await atom.workspace.open("sample.js");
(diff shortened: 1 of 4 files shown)


dirk-thomas:
@maxbrunsfeld The current state of master seems to fail the builds against unstable on Windows (this PR only adds a few comment lines at the moment). Before I can iterate on this it would be good if unstable passes again. Can you please take care of that?

dirk-thomas:
@maxbrunsfeld Any news on fixing the Windows CI for the beta?

50Wliu:
Tada!

dirk-thomas:
Great. I will look into updating the specs to cover the new context lines.

spec/results-model-spec.js
```diff
@@ -33,6 +35,10 @@ describe("ResultsModel", () => {
       expect(resultsModel.getPathCount()).toBe(1);
       expect(resultsModel.getMatchCount()).toBe(6);
       expect(resultsModel.getPaths()).toEqual([editor.getPath()]);
+      if (parseFloat(atom.getVersion()) >= 1.17) {
```
dirk-thomas: For the [previous PR](https://github.com/atom/find-and-replace/pull/847/files#diff-eb24277759b68118e9e07ef629d193e8R250) the following was recommended to compare the Atom version: https://github.com/atom/find-and-replace/pull/847#issuecomment-293036002 Looking back the interpretation as a float is just wrong. E.g. version 1.2 would satisfy this check. Since I need this for the specs too I wanted to ask what I should use instead. Is there a recommended way to compare the version number? Should I pull in `semver` as  a direct dependency of this package?

50Wliu: Considering that this check is going to be short-term and only for specs I think it's safe to keep the float interpretation.

dirk-thomas: Please see the first reference. The same condition has been added to the code of this package. Since the package doesn't specify a minimum version I think it should be changed.

50Wliu: I still think it wouldn't matter given that find-and-replace is a core package and is tied to specific Atom versions.

dirk-thomas:
How does this look to verify that the correct context lines are present in the model as well as the view?

spec/results-view-spec.js
```diff
@@ -701,6 +701,32 @@ describe('ResultsView', () => {
...
+
+    it('shows the context lines', async () => {
+      if (parseFloat(atom.getVersion()) >= 1.17) {
+        const pathNode = resultsView.refs.listView.element.querySelectorAll(".search-result")[0];
+        expect(pathNode.querySelectorAll('.list-tree .list-item').length).toBe(5)
+        expect(pathNode.querySelectorAll('.list-tree .list-item')[0]).not.toHaveClass('match-line');
+        expect(pathNode.querySelectorAll('.list-tree .list-item')[1]).toHaveClass('match-line');
+        expect(pathNode.querySelectorAll('.list-tree .list-item')[2]).not.toHaveClass('match-line');
+        expect(pathNode.querySelectorAll('.list-tree .list-item')[3]).not.toHaveClass('match-line');
+        expect(pathNode.querySelectorAll('.list-tree .list-item')[4]).not.toHaveClass('match-line');
```
maxbrunsfeld: Could you also assert about each list item's text content? With so many assertions, it might be good to extract the `querySelectorAll` expression into a variable, like `listItems` or something.

dirk-thomas: Done in 84b5e07121a1902fc4da2440b81ed27a13e848ec

spec/results-view-spec.js
```diff
@@ -701,6 +701,41 @@ describe('ResultsView', () => {
...
+        expect(pathNodes.length).toBe(1)
+        const pathNameNode = pathNodes[0].querySelector('.path-name');
+        expect(pathNameNode.textContent).toBe('sample.coffee');
+        const resultNode = pathNodes[0].querySelector('.search-result');
+        const lineNodes = resultNode.querySelectorAll('.list-item');
+        expect(lineNodes.length).toBe(5)
+        expect(lineNodes[0]).not.toHaveClass('match-line');
+        expect(lineNodes[0].querySelector('.preview').textContent).toBe('class quicksort');
+        expect(lineNodes[1]).toHaveClass('match-line');
+        expect(lineNodes[1].querySelector('.preview').textContent).toBe('sort: (items) ->');
```
dirk-thomas: The matched line currently doesn't render leading whitespace. But the context line currently do. Should I strip the leading whitespaces for the context lines in the view?

maxbrunsfeld: Ah, thanks for checking. Yeah, I don't have a strong opinion on this, but I'd say let's strip whitespace just to be consistent w/ the existing match line.

dirk-thomas: Done in 0c614b484e9b723352d81f6a2b451fe678132e38

maxbrunsfeld:
Thanks @dirk-thomas!

dirk-thomas:
I am looking forward to improve the view with additional features. Some things I have in my mind which would be really helpful:

* Fold/unfold the context lines
* Merge redundant lines

When I find time I will start a discussion how that could look like...

50Wliu:
A bit late, but [trimLeft](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trimLeft) seems to be non-standard.  Perhaps something else should be used?

dirk-thomas:
It was already used for the match line. Therefore I just used the same for the context lines.

I will likely create a pull request soon which will remove the trimming of leading spaces anyway (since I think they are essential and might contain information useful to the user to decide which search result they want to look at).

dirk-thomas:
@maxbrunsfeld Will this be released and used in Atom master so that it will be included in time for 1.18 beta?

dirk-thomas:
@maxbrunsfeld Imo this change should be released for 1.18 since it fixes the leading spaces.

Issue #683, opened 2016-03-17
Find panel close icon


depperm:
There have been previous issues that mention this (#164, #215, #219, #201, #204) and a few related ones about it not closing with ctrl-f. I believe a GUI close button could be added by modifying find-view.coffee
Adding lines about:
line 32: `@span class: 'icon icon-x pull-right', style: 'padding-left:5px', outlet: 'closeFindPanel'`
line 156: `@handleCloseEvents()`
line 192: `handleCloseEvents: ->`
`@closeFindPanel.on 'click', (e) => @panel?.hide()`

lee-dohm:
Thanks very much for the feedback. We've listened to the requests and we're not interested in adding a close button to the find-and-replace panel.

Issue #219, opened 2014-05-21
Panel needs a close button / Cannot close find and replace


harrypujols:
How do I close the panel? It needs a :negative_squared_cross_mark: button somewhere. Even if the key shortcut is somewhere around (I can't find it), it's a good UI practice to have a visual option.

Issue #204, opened 2014-05-12
Add close button


lfuelling:
I just can't figure out a way to close the find and replace dialog without restarting the editor. It would be awesome, if a "close" button could be added.

![2014-05-12--1399887597_1396x142_scrot](https://cloud.githubusercontent.com/assets/5144843/2942774/812534ca-d9b9-11e3-9580-88920481b9f7.png)

PR #885, opened 2017-04-18
add close button

UziTech:
### Description of the Change

Add a close button to the Find in Current Buffer panel and the Find in Project panel

image

### Alternate Designs

Could be on the left

### Benefits

People with a 2016 Macbook pro can finally close the find-and-replace panel

### Possible Drawbacks

Takes up screen real estate. Might need to consider a setting to enable/disable it.

### Applicable Issues

#164
#204
#219
#347
#683

### DIFF
--- a/lib/find-view.js
+++ b/lib/find-view.js
@@ -469,7 +469,7 @@ class FindView {
 
   clearMessage() {
     this.element.classList.remove('has-results', 'has-no-results');
-    this.refs.descriptionLabel.innerHTML = 'Find in Current Buffer <span class="subtle-info-message">Close this panel with the <span class="highlight">esc</span> key</span>'
+    this.refs.descriptionLabel.innerHTML = 'Find in Current Buffer'
     this.refs.descriptionLabel.classList.remove('text-error');
   }
 
--- a/lib/find-view.js
+++ b/lib/find-view.js
@@ -31,6 +31,10 @@ class FindView {
     return (
       $.div({tabIndex: -1, className: 'find-and-replace'},
         $.header({className: 'header'},
+          $.span({ref: 'closeButton', className: 'header-item close-button pull-right'},
+            $.i({className: "icon icon-x clickable"})
+          ),
+
           $.span({ref: 'descriptionLabel', className: 'header-item description'},
             'Find in Current Buffer'
           ),
@@ -164,6 +168,10 @@ class FindView {
     if (this.tooltipSubscriptions) return;
 
     this.tooltipSubscriptions = new CompositeDisposable(
+      atom.tooltips.add(this.refs.closeButton, {
+        title: 'Close Panel <span class="keystroke">Esc</span>',
+        html: true
+      }),
       atom.tooltips.add(this.refs.regexOptionButton, {
         title: "Use Regex",
         keyBindingCommand: 'find-and-replace:toggle-regex-option',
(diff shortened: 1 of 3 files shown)


reggieb:
+1 to this PR

thomasjo:
> People with a 2016 Macbook pro can finally close the find-and-replace panel

This is still possible, although does require an extra keypress to reveal the function keys.

> Takes up screen real estate. Might need to consider a setting to enable/disable it.

**No**. We are not adding a setting to enable/disable a UI element.

While I'm not opposed to the notion of having a close button for the find and replace panel, from the screenshot it is obvious that the current implementation/style breaks the visual design of the panel. The current implementation therefore needs to be reworked. Additionally, this likely needs to be implemented at the core UI level. In other words, support for panel close buttons should perhaps be implemented in [atom/atom-ui](https://github.com/atom/atom-ui) first, then this PR should merely leverage the core UI functionality.

/cc @simurai @atom/feedback

simurai:
Now that there are docks https://github.com/atom/atom/pull/13977, my ✋  vote would go towards moving find-and-replace into the bottom dock. Here a quick prototype:

find

Then the space gets shared with other bottom panels, in the above example it's Linter. You can still keep using the keyboard like you're used to, but when hovering the tab, there will be a close button. Some themes show it all the time. For people that are REALLY concerned about the extra space for the tabs could still hide them in their `styles.less` file.

Another bonus: You then even could move find-and-replace to the left or right dock. Maybe above the search results? Although that might need some style tweaks to make it look good.

ScreamingHawk:
Just started using atom and this was the first problem I ran into. Having to look up bug report to find a hidden keyboard shortcut to do something this trivial, leaves a bad taste. Noticed people have had problems finding this since 2014 and it hasn't been addressed. 

Coming from Notepad++, IntelliJ (and derivatives)

UziTech:
@ScreamingHawk Welcome to Atom. In the mean time you can install my package [find-and-replace-close-button](https://atom.io/packages/find-and-replace-close-button) until they have this worked out.

jeffmcneill:
According to the above, the pr is working in the stable version, but broken in the beta version. Does that mean beta is broken or the pr code?

50Wliu:
Beta was broken when this PR was created (it looked in the wrong place for bundled packages).  Going to restart the build so that it uses the latest Beta version, which should fix it.

simurai:
> my ✋  vote would go towards moving find-and-replace into the bottom dock.

It looks like find-and-replace will **not** move to the bottom dock and stay its own panel.

So here two tweaks: https://github.com/UziTech/find-and-replace/pull/1

Otherwise I'm 👍  on merging.

> @thomasjo No. We are not adding a setting to enable/disable a UI element.

Also, ok with that, since hiding the close button could also be a done with `.find-and-replace .close-button { display: none; }` in your `styles.less`. For those that _really_ don't like it.

UziTech:
I removed the commit for the enable/disable setting and added the changes in https://github.com/UziTech/find-and-replace/pull/1

maxbrunsfeld:
Thanks @UziTech! Nice work.

simurai:
Ok, so for people that really ~hate~ dislike the close button, you can add this line to your `styles.less` and it should disappear:

```less
// hide find+replace close button
.find-and-replace .close-button { display: none; }
```

PR #886, opened 2017-04-18
Make tests compatible with new editor rendering

nathansobo:
Just a couple tweaks so we can get this building on https://github.com/atom/atom/pull/13880. It should still build green against stable.


nathansobo:
Failures on Windows against beta seem unrelated.

Issue #592, opened 2015-11-09
Trigger project-find:show-in-current-directory with keymap


timkelty:
I want the "Search in directory" functionality to be the default for the `shift-cmd-F` keymap.

This triggers the find in project panel (and I confirmed it is triggering my keymap.cson command), but the path isn't populated with the current directory.

```
'.platform-darwin':
  'shift-cmd-F': 'project-find:show-in-current-directory'
```

jaketrent:
+1 to this desire.  Works well in nuclide-file-tree, but I'd like to use atom/tree-view instead at this point.

timkelty:
Looks like a duplicate of #276

PR #888, opened 2017-04-22
Trigger project-find:show-in-current-directory with hotkey

suside:
### Description of the Change
When `project-find:show-in-current-directory` is triggered with a mouse click the element we get is span, therefore searching for ancestor with `closest('.directory')` works.
But when it's triggered with a hotkey we get main `ol .tree-view`, thus we need to search for selected directory in descendants.
### Alternate Designs
None that i can think of.
### Benefits
Bug fixed :)
### Possible Drawbacks
None that i can think of.
### Applicable Issues
#592

### DIFF
--- a/lib/project-find-view.js
+++ b/lib/project-find-view.js
@@ -354,7 +354,7 @@ class ProjectFindView {
   }
 
   directoryPathForElement(element) {
-    const directoryElement = element.closest('.directory');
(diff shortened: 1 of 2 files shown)


suside:
well... whoever reads this.. should switch to vscode by now :crossed_fingers:

PR #905, opened 2017-05-29
Change "Search in Directory" to "Search in Folder"

aninternetof:
### Description of the Change

There is some inconsistency between us and [tree-view](https://github.com/atom/tree-view/blob/master/menus/tree-view.cson) in choosing the term "folder" or "directory" in the context menu. The "Search in Directory" context menu item is shown alongside tree-view's "New Folder" item. This PR changes our item to "Search in Folder" for consistency.

screenshot

### Alternate Designs

Alternatively, we could have tree-view change its "Folder" items to say "Directory". But they have several, and we have one.

### Benefits

It makes the resulting context menu more pleasing to read.

### Possible Drawbacks

It looks like most of the internal code uses the word "directory", so the user-visible label will not match the internals, which is always unfortunate.

### Applicable Issues

None.

### DIFF
--- a/menus/find-and-replace.cson
+++ b/menus/find-and-replace.cson
@@ -23,7 +23,7 @@
 
 'context-menu':
   '.tree-view li.directory': [
-    { 'label': 'Search in Directory', 'command': 'project-find:show-in-current-directory' }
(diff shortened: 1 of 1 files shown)


daviwil:
Thanks @aninternetof!  I agree, this is more consistent with the other items in this context menu.

PR #908, opened 2017-06-03
Toggle context lines

dirk-thomas:
### Description of the Change

The package has config settings for the number of context lines before and above each matched line. The default values are zero in order to not change previous behavior (and show no context by default). When any of these values is set to non-zero the view always shows the context lines.

This patch adds a button to the results pane to toggle between showing and hiding the context lines. If the config values for the context lines are both zero the button is not shown since it wouldn't do anything.

### Benefits

The user can toggle the context lines without the need to change the package settings and redo the query.

### Possible Drawbacks

Currently there is no option to hide the context lines by default (when non-zero values are in the config). Maybe the latest state can be made persistent to use it for the next search result?

### Applicable Issues

* The scroll position doesn't stay around the same matches when the context lines are being toggled.

### Future changes

As a consequence I think it would make sense to change the default values of the config from zero to e.g. three. If the context lines are hidden by default (and after being toggled remember the last state) that would maintain the previous behavior but make the context available to all users by default without requiring custom settings.

### DIFF
--- a/lib/find-options.coffee
+++ b/lib/find-options.coffee
@@ -9,6 +9,7 @@ Params = [
   'wholeWord'
   'caseSensitive'
   'inCurrentSelection'
+  'showContextLines'
 ]
 
 module.exports =
@@ -23,6 +24,7 @@ class FindOptions
     @caseSensitive = state.caseSensitive ? atom.config.get('find-and-replace.caseSensitive') ? false
     @wholeWord = state.wholeWord ? atom.config.get('find-and-replace.wholeWord') ? false
     @inCurrentSelection = state.inCurrentSelection ? atom.config.get('find-and-replace.inCurrentSelection') ? false
+    @showContextLines = state.showContextLines ? atom.config.get('find-and-replace.showContextLines') ? false
 
   onDidChange: (callback) ->
     @emitter.on('did-change', callback)
--- a/lib/find-options.coffee
+++ b/lib/find-options.coffee
@@ -9,7 +9,8 @@ Params = [
   'wholeWord'
   'caseSensitive'
   'inCurrentSelection'
-  'showContextLines'
+  'leadingContextLineCount'
+  'trailingContextLineCount'
 ]
 
 module.exports =
@@ -24,7 +25,8 @@ class FindOptions
     @caseSensitive = state.caseSensitive ? atom.config.get('find-and-replace.caseSensitive') ? false
     @wholeWord = state.wholeWord ? atom.config.get('find-and-replace.wholeWord') ? false
     @inCurrentSelection = state.inCurrentSelection ? atom.config.get('find-and-replace.inCurrentSelection') ? false
-    @showContextLines = state.showContextLines ? atom.config.get('find-and-replace.showContextLines') ? false
+    @leadingContextLineCount = state.leadingContextLineCount ? atom.config.get('find-and-replace.leadingContextLineCount') ? 0
+    @trailingContextLineCount = state.trailingContextLineCount ? atom.config.get('find-and-replace.trailingContextLineCount') ? 0
 
   onDidChange: (callback) ->
     @emitter.on('did-change', callback)
--- a/lib/find-view.js
+++ b/lib/find-view.js
@@ -145,7 +145,7 @@ class FindView {
             <rect x="3" y="6" width="8" height="2"></rect>
             <rect x="3" y="9" width="8" height="2"></rect>
             <path d="M12,3 L15,3 L15,11 L12,11 L12,10 L14,10 L14,4 L12,4  Z"></path>
-          </svg>
+          </symbol>
 
           <symbol id="find-and-replace-context-lines-after" viewBox="0 0 20 16" stroke="none" fill-rule="evenodd">
             <rect opacity="0.6" x="3" y="3" width="14" height="2"></rect>
@@ -153,7 +153,7 @@ class FindView {
             <rect x="3" y="9" width="8" height="2"></rect>
             <rect x="3" y="12" width="8" height="2"></rect>
             <path d="M12,6 L15,6 L15,14 L12,14 L12,13 L14,13 L14,7 L12,7  Z"></path>
-          </svg>
+          </symbol>
         `})
       )
     );
--- a/lib/find-view.js
+++ b/lib/find-view.js
@@ -140,19 +140,15 @@ class FindView {
           </symbol>
 
           <symbol id="find-and-replace-context-lines-before" viewBox="0 0 20 16" stroke="none" fill-rule="evenodd">
-            <rect opacity="0.6" x="3" y="12" width="14" height="2"></rect>
-            <rect x="3" y="3" width="8" height="2"></rect>
-            <rect x="3" y="6" width="8" height="2"></rect>
-            <rect x="3" y="9" width="8" height="2"></rect>
-            <path d="M12,3 L15,3 L15,11 L12,11 L12,10 L14,10 L14,4 L12,4  Z"></path>
+            <rect opacity="0.6" x="2" y="11" width="16" height="2"></rect>
+            <rect x="2" y="7" width="10" height="2"></rect>
+            <rect x="2" y="3" width="10" height="2"></rect>
           </symbol>
 
           <symbol id="find-and-replace-context-lines-after" viewBox="0 0 20 16" stroke="none" fill-rule="evenodd">
-            <rect opacity="0.6" x="3" y="3" width="14" height="2"></rect>
-            <rect x="3" y="6" width="8" height="2"></rect>
-            <rect x="3" y="9" width="8" height="2"></rect>
-            <rect x="3" y="12" width="8" height="2"></rect>
-            <path d="M12,6 L15,6 L15,14 L12,14 L12,13 L14,13 L14,7 L12,7  Z"></path>
+            <rect x="2" y="11" width="10" height="2"></rect>
+            <rect x="2" y="7" width="10" height="2"></rect>
+            <rect opacity="0.6" x="2" y="3" width="16" height="2"></rect>
           </symbol>
         `})
       )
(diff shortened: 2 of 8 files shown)


dirk-thomas:
Should the state for showing / hiding the context lines be added to `FindOptions` which would make the setting persistent?

iolsen:
@maxbrunsfeld do you have some time to review this?

maxbrunsfeld:
@dirk-thomas Can you post a GIF of the interaction you've got working so far?

@simurai What direction would you like to see for this feature?

I could actually imagine a UI element that raises and lowers the number of context lines, like little `+` and `-` buttons with the word `context` between them.

dirk-thomas:
Please see the following gif which uses the config settings of 3 lines of context before and after. Sorry for the bad resolution - I couldn't make the screen recorder work with higher resolutions:
find-and-replace_toggle-context-lines

I would be interested to make the following changes (in that order) if there is interest in that and pull request will be reviewed in a reasonable time frame:

* update this PR:
  * to persist the toggle state, e.g. in the find options (see my pending question)
  * maintain a reasonable scrolling position when toggling (whatever that means)
* separate PRs:
  * change the default setting to N context lines (e.g. 3) so that users can toggle the context lines without requiring custom settings
  * render leading whitespace correctly (if necessary add an option to enable/disable this behavior)
  * deduplicate lines in the search results (duplicate context lines as well as multiple matches per line)

> I could actually imagine a UI element that raises and lowers the number of context lines, like little `+` and `-` buttons with the word `context` between them.

That would be an interesting feature to. The `+` command will currently be constraint by the number of context lines configure in the settings (since only that many context lines are being fetched). Otherwise when that threshold is exceeded the search query would need to be repeated with different arguments.

simurai:
> update this PR:
> - to persist the toggle state, e.g. in the find options (see my pending question)
> - maintain a reasonable scrolling position when toggling (whatever that means)

👍  I think that would be great.

Plus maybe move the button in its own `btn-group` so that it feels more like a separate toggle and not belonging to the collapse buttons. I can help out with that if you run into styling troubles.

screen shot 2017-06-15 at 5 05 58 pm

And change to a "Show Context" label. So it starts with a verb.

---

> I could actually imagine a UI element that raises and lowers the number of context lines, like little `+` and `-` buttons with the word `context` between them.

Yeah, that would be similar to the Image View zoom. Although not sure if it's annoying having to click multiple times and you can't just expand straight to your prefered amount of lines. I think if you wanna seem more, you wanna see as much as possible (as set in the config).

---

Or another option: Expand context lines per match. Usually when I want to see more context it's not for every match, but just the one that I might be interested in. Maybe have an expand button and/or a keyboard shortcut like `cmd-enter` to toggle context only for that match.

screen shot 2017-06-15 at 5 04 44 pm

Anyways, that is outside of this PR's scope.

dirk-thomas:
@simurai I moved the new button into a separate group, left of the existing buttons and updated the label (8ea44091216f0c55366933e9db1a137997a43576). I had to use a wrapping `btn-toolbar` to visually separate the two groups. Was that how you imagined it?

find-and-replace_update-button

The per-match buttons might offer more flexibility but I would expect that it would still require a button to toggle the context lines for all matches for convenience. Having both might also make the UI more crowded than necessary. So I kept it with a single "global" button for now.

dirk-thomas:
I still kind of need answers / comments regarding the two pending issue:

* to persist the toggle state: Should I use the find options for this?
* maintain a reasonable scrolling position when toggling: What exactly should than achieve? The first visible match should stay close to the top? Or any other criteria?

simurai:
> maintain a reasonable scrolling position when toggling: What exactly should than achieve? The first visible match should stay close to the top? Or any other criteria?

Hmm.. This is probably super hard to get right. If you're scrolled down a bit and then click on "Show Context" and it jumps to the first match, you would have to scroll back to the position that you were before. Would some sort of "keep it centered" be possible?

maxbrunsfeld:
Storing the toggle state in the find options is fine.

Also, we probably shouldn't show this button if both `searchContextLineCountBefore` and `searchContextLineCountAfter` are set to zero, since it won't do anything.

dirk-thomas:
> Storing the toggle state in the find options is fine.

I will try to update the patch with this soon.

> Also, we probably shouldn't show this button if both `searchContextLineCountBefore` and `searchContextLineCountAfter` are set to zero, since it won't do anything.

This is already the case in the current patch.

dirk-thomas:
See edf279fe2a9d851e8bce02eaf2915f967c7fe16c for making the state to show context lines persistent via the FindOptions.

@simurai Regarding the idea of `-` and `+` buttons: for some use cases I can see that it would be nice to e.g. only show N lines of context *after* the match (but not any before). Also what label should be used for in-between the `-` and `+` buttons? It should probably indicate the number of currently shown lines as well as mention "Context" somehow. If we use two separate parts for before and after the label should also clarify which one it belongs to. If the user e.g. configures `0` as the max lines before but `N > 0` as the maximum lines after only the later controls will be visible and should be clear enough (even when the controls for the previous lines are not around). Maybe create two new icons (for context lines before and after) would be best and the label would show those icons as well as the currently used number... So many ideas & options :wink: Any input on the UI part would be greatly appreciated - especially when considering custom icons I would probably need help from someone creating those...

dirk-thomas:
Regarding the scrolling behavior: the existing collapse / expand option to hide/fold and show/unfold the matches per path avoids the same problem by simply always scrolling to the top when one of the two options is selected. I would say this would be a "good enough" default behavior for the new show/hide context lines options too.

If we figure out a decent way how to behave in that case (which is reasonable to implement and doesn't lead to weird behavior in corner cases) it can be applied to all the options where the lengths of the search results view changes.

Does that sound like a viable option? I have implemented this behavior in 41faed0c7ab1e48e81989882eed59f2dbeacfb16 for now.

dirk-thomas:
I also implemented the separate `+` and `-` buttons for leading as well as trailing context lines. Please the the following animation for how it looks and works (note the flickering in the beginning is only visible in the screen recording):

find-and-replace_dec-inc-sep-context-lines

I would replace the words "before" and "after" with an icon. Maybe even the line count could be removed from that button since it should be obvious from the results visible below.

What do you think?

maxbrunsfeld:
Wow, that looks really cool @dirk-thomas, and the code looks great so far!

> the existing collapse / expand commands to hide/fold and show/unfold the matches per path avoid the same problem by simply always scrolling to the top when the commands are triggered. I would say this would be a "good enough" default behavior for the new show/hide context lines behavior too.

I think scrolling to the top would be a bit more jarring for these options, because if you're clicking these buttons, there's probably a particular search result that you're looking at and wanting to see context for. What do you think about at least scrolling such that the same *file* is visible?

I'm going to defer to @simurai on some of the other UI design questions.

dirk-thomas:
Please see the [latest commit](7f89f9818757c8019b5e358bf0a4875ceda08bdd) using custom made icons instead:

find-and-replace_context-icons

dirk-thomas:
> I think scrolling to the top would be a bit more jarring for these options, because if you're clicking these buttons, there's probably a particular search result that you're looking at and wanting to see context for. What do you think about at least scrolling such that the same file is visible?

Well, I think "the same file is visible" is already a bit too fuzzy. Multiple files might be in the viewport. Which one should be picked in that case? I think it will be pretty complicated to handle this (however the intended behavior should be). So maybe keep this for a separate patch?

maxbrunsfeld:
> Well, I think "the same file is visible" is already a bit too fuzzy. Multiple files might be in the viewport. Which one should be picked in that case?

> I think it will be pretty complicated to handle this

Ok sorry, I thought maybe leaving it semi-open ended would be helpful. How about this: after pushing one of those buttons, we should scroll to reveal the selected match. It shouldn't be complicated; similar functionality is already used for most of the existing commands. See [the scrollToSelectedMatch method](https://github.com/atom/find-and-replace/blob/fc6b4c3b08a3aba9bac39880e2701f5706744f1f/lib/project/results-view.js#L443-L454).

dirk-thomas:
> How about this: after pushing one of those buttons, we should scroll to reveal the selected match.

The downside of that heuristic is that the currently selected match might not even be in the visible area before the user clicks the button.

maxbrunsfeld:
That's true, but I think it's a reasonable behavior, and it's not obvious that it's any less desirable than scrolling to the currently visible match. It's also slightly easier, so I'm inclined to start with that.

dirk-thomas:
> ... scrolling to the currently visible match

Changed in ddd2cf3645dd3d7c6073f0dcb0bfaa5c2af2baf9.

dirk-thomas:
I just found one problem which got introduced when I added the icons to the toggle buttons: I can only click the buttons on the "sides" (beside the icons). If I try to click the buttons in the middle (where the icons are) the clicked event doesn't fire. I must be doing something wrong with the icon. I found a wrong closing tag for my icons (fixed in 09fbc22bc1d73129c3bfbd4238746d5f13b9ca35) but the problem is still the same. I don't see what is different compared to the other icon. Any ideas?

simurai:
> I can only click the buttons on the "sides" (beside the icons).

Should be fixed. Not sure why it works for the other icons. Maybe those buttons use some other event bubbling trick.

dirk-thomas:
@simurai Thank you for the fix. I can confirm it works for me too.

@maxbrunsfeld Ready for another round of review I think.

maxbrunsfeld:
@dirk-thomas The code looks very good. It looks like `incrementLeadingContextLines` and friends are not yet tested. Could you add another test that exercises these four methods? Maybe it could do a search with only one match so that it's really easy to reason about what matches are on-screen.

Other than that, I think I'm ready to merge this. Thanks so much for continuing this whole effort, I think it's now going to really make an impact.

dirk-thomas:
> Could you add another test that exercises these four methods?

I added more specs to use all the context lines function to increase/toogle/decrease leading/trailing lines: 61c09516c1185260a64e6068d29d0e1e2712d055

> Maybe it could do a search with only one match so that it's really easy to reason about what matches are on-screen.

I created a function to extract the first match which gets rid of duplicate code. I still checked that the "right" lines are shown though and didn't rely only on the count which would be possible if there is a single match.

dirk-thomas:
@maxbrunsfeld Any update on reviewing / merging this?

maxbrunsfeld:
This is awesome.

maxbrunsfeld:
@dirk-thomas This has been incorporated into Atom master and will go out in Atom 1.20.

dirk-thomas:
Great, thank you. I already have the next patch ready for you: #919 :wink:

dirk-thomas:
Please see #920 which contains a fix for a regression introduced in this patch.

PR #919, opened 2017-06-30
keep leading whitespace

dirk-thomas:
### Description of the Change

Currently each line shown in the result view has the leading spaces stripped. For some context / languages whitespace is critical. Especially when also showing context line the relative indentation can be very important.

This patch keep the leading whitespace and makes sure it is rendered correctly (not collapsing multiple spaces into a single visual space).

### Alternate Designs

If the new behavior is not considered desired an option could be introduced to toggle the two possible behaviors (hide leading whitespace vs. showing them).

### Benefits

Context awareness.

### Possible Drawbacks

The additional indentation will move the text of the line further to the right.

### Applicable Issues

None, as far as I know.

### DIFF
--- a/lib/project/match-view.js
+++ b/lib/project/match-view.js
@@ -45,7 +45,7 @@ class MatchView {
     const range = Range.fromObject(this.match.range);
     const matchStart = range.start.column - this.match.lineTextOffset;
     const matchEnd = range.end.column - this.match.lineTextOffset;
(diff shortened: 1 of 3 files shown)


dirk-thomas:
@maxbrunsfeld What do you think about this patch?

dirk-thomas:
Thanks for merging this. Is it possible to make a release for this in order to be included into Atom 1.20?

PR #920, opened 2017-07-01
increase default number of queries context lines

dirk-thomas:
### Description of the Change

The patch increases the default values of queried context lines.

**Update:** this identified a regression in the scrolling behavior which has been fixed in the second commit. More details in the next comment.

### Alternate Designs

Every user would need to change the settings to benefit from the new functionality to show context lines around search results. With the default of zero non of these UI improvements implemented in #908 are available.

### Benefits

While the queried context lines are not shown by default the increased default values will make the UI elements to show context lines around search results on demand available to all users.

### Possible Drawbacks

Slightly more memory usage for search results since they store the additionally queried context lines.

### Applicable Issues

None afaik.

### DIFF
--- a/lib/project/results-view.js
+++ b/lib/project/results-view.js
@@ -264,13 +264,22 @@ class ResultsView {
       for (let j = 0; j < this.selectedMatchIndex; j++) {
         const match = selectedResult.matches[j];
         result += this.matchHeight;
-        if (match.leadingContextLines) result += this.contextLineHeight * match.leadingContextLines.length;
-        if (match.trailingContextLines) result += this.contextLineHeight * match.trailingContextLines.length;
+        if (match.leadingContextLines && this.model.getFindOptions().leadingContextLineCount) {
+          result += this.contextLineHeight * Math.min(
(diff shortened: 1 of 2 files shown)


dirk-thomas:
The second commit fixes a regression of #908. When the number of context lines queried (the settings) is different from the number of lines the user has selected to show (the -/+ UI controls) the height for scrolling was calculated incorrectly. It was always using the settings. With this patch the minimum of both values is being used to calculate the actual height of the visual elements correctly.

@maxbrunsfeld You mentioned that #908 was merged and released for the upcoming Atom 1.20. I think this (at least the second commit) is an important fix and should be merged and released as well. Otherwise the scrolling behavior will have a regression when users change the default settings to any number of context lines > 0.

maxbrunsfeld:
📈 Great work, @dirk-thomas. Now everyone will start enjoying this feature that you've spearheaded.

maxbrunsfeld:
This is on Atom master.

dirk-thomas:
Thanks!

PR #922, opened 2017-07-03
Enable autocompletion in the find editor

wvanlint:
### Description of the Change

This pull request is a continuation of #710 after addressing comments by @as-cii.

> This PR enables autocompletion in the find editor by:
> - Registering the find editor so it's observable by autocomplete-plus
> - Adding a corresponding CSS selector to the default provider so it knows to provide completions for that editor.

Depends on atom/autocomplete-plus#864.

### Benefits

This will provide autocompletion when searching.

### Applicable Issues

Fixes atom/find-and-replace#92.

### DIFF
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -122,6 +122,20 @@ module.exports =
     new Disposable ->
       FileIcons.resetService()
 
+  requestAutocompletions: ->
+    disposable = @autocompleteWatchEditor?(@findView.findEditor, ['default'])
+    if disposable?
+      @autocompleteSubscriptions.add(disposable)
+
+  consumeAutocompleteWatchEditor: (watchEditor) ->
+    @autocompleteSubscriptions = new CompositeDisposable
+    @autocompleteWatchEditor = watchEditor
+    if @findView?
+      requestAutocompletions()
+    new Disposable =>
+      @autocompleteSubscriptions.dispose()
+      @autocompleteWatchEditor = null
+
   provideService: ->
     resultsMarkerLayerForTextEditor: @findModel.resultsMarkerLayerForTextEditor.bind(@findModel)
 
@@ -139,6 +153,7 @@ module.exports =
     options = {findBuffer, replaceBuffer, pathsBuffer, findHistoryCycler, replaceHistoryCycler, pathsHistoryCycler}
 
     @findView = new FindView(@findModel, options)
+
     @projectFindView = new ProjectFindView(@resultsModel, options)
 
     @findPanel = atom.workspace.addBottomPanel(item: @findView, visible: false, className: 'tool-panel panel-bottom')
@@ -160,6 +175,8 @@ module.exports =
     # See https://github.com/atom/find-and-replace/issues/63
     ResultsPaneView.model = @resultsModel
 
+    @requestAutocompletions()
+
   deactivate: ->
     @findPanel?.destroy()
     @findPanel = null
@@ -176,6 +193,8 @@ module.exports =
     ResultsPaneView.model = null
     @resultsModel = null
 
+    @autocompleteSubscriptions?.dispose()
+    @autocompleteManagerService = null
     @subscriptions?.dispose()
     @subscriptions = null
 
--- a/package.json
+++ b/package.json
@@ -45,6 +45,11 @@
       "versions": {
         "1.0.0": "consumeFileIcons"
       }
+    },
+    "autocomplete.watchEditor": {
+      "versions": {
+        "1.0.0": "consumeAutocompleteWatchEditor"
+      }
     }
   },
   "providedServices": {


lib/find-view.js
```diff
@@ -23,6 +23,11 @@ class FindView {
     this.updateOptionViews();
     this.updateReplaceEnablement();
     this.createWrapIcon();
+
+    if (atom.textEditors) {
+      var disposable = atom.textEditors.add(this.findEditor);
```
nathansobo: Should we add the replace editor as well? Also, what about `ProjectFindView`?

lib/find.coffee
```diff
@@ -117,6 +117,20 @@ module.exports =
       'find-and-replace:select-skip': (event) ->
         selectNextObjectForEditorElement(this).skipCurrentSelection()
 
+    maybeRegisterToAutocompletePlus = (atomPackage) =>
```
nathansobo: It would be better to expose a service from autocomplete plus that exposes the ability to observe providers and consume it in this package.

But another question... what do you think about hard-coding support for find-and-replace editors directly in the default autocomplete provider rather than adding it dynamically? Then you'd only need to register the editor in this package.

lib/find-view.js
```diff
@@ -23,6 +23,11 @@ class FindView {
...
+      this.findEditor.onDidDestroy(() => disposable.dispose());
```
nathansobo: I don't think we destroy these editors when the package deactivates, so I'm not sure this will work as intended. Since these editors live for the entire time the package is active and we never create more than one find panel, it might be better to just dispose the registration when this package deactivates.

lib/find-view.js
wvanlint: I thought about adding autocompletion to the replace editor too, but the use case is not as clear as it is less likely that the replacement will be an existing symbol. I can add it if you'd like though.

As for `ProjectFindView`, none of the providers are set up to provide autocompletions from an entire project. They only provide completions based on (active) pane items. I definitely would like to look into this later, but for now, I don't think it would be possible to enable this.

lib/find-view.js
wvanlint: Done. I can also destroy the editors in `FindView` if that's preferable.

lib/find.coffee
wvanlint: > It would be better to expose a service from autocomplete plus that exposes the ability to observe providers and consume it in this package.

I might be missing something, but isn't that similar to this implementation, only that here the `ProviderManager` exposes the ability? Or would you like this to avoid the coupling with the `ProviderManager`?

About hard-coding the support, that definitely makes sense too! I only thought it might be beneficial to have a recipe for adding autocompletion to X, where X might be a third-party package.

nathansobo: I'm mainly concerned about interacting with the package registry in this way. Services are a cleaner way of interacting with another package and are also semantically versioned, which is important as APIs evolve.

But I think I'd prefer to make the change in `autocomplete-plus` directly. It seems simpler on balance. I wonder if you could craft a generic class like `provide-default-autocompletions` or something that anyone could apply to an editor they register?

nathansobo:
I'm going to wait to review this until we adjust the service interface to be synchronous pending https://github.com/atom/autocomplete-plus/pull/876.

lib/find.coffee
```diff
@@ -122,6 +122,20 @@ module.exports =
...
+  requestAutocompletions: (editor, labels) ->
+    disposable = @autocompleteWatchEditor?(editor, labels)
+    if disposable?
+      @autocompleteSubscriptions.add(disposable)
+
+  consumeAutocompleteWatchEditor: (watchEditor) ->
+    @autocompleteSubscriptions = new CompositeDisposable
+    @autocompleteWatchEditor = watchEditor
+    if @findView?.findEditor?
+      requestAutocompletions(@findView.findEditor, ['default'])
```
wvanlint: Hmm, I think the client libraries (here `find-and-replace`) should choose which providers are applicable, instead of the other way around.

Maybe some more descriptive labels will be helpful later on, such as `active-editor-symbols` or `project-symbols` etc.

lib/find.coffee
```diff
@@ -139,6 +153,8 @@ module.exports =
     options = {findBuffer, replaceBuffer, pathsBuffer, findHistoryCycler, replaceHistoryCycler, pathsHistoryCycler}
 
     @findView = new FindView(@findModel, options)
+    @requestAutocompletions(@findView.findEditor, ['default'])
```
wvanlint: Done.

nathansobo:
Sorry for the delay in merging this @wvanlint. This release has been a bumpy one with all the changes we made.

nathansobo:
Hey. I just merged this and it doesn't seem to be working. Have you tried it of late?

wvanlint:
Thanks for the merge!
I just checked again with the HEAD versions of Atom, `find-and-replace` and `autocomplete-plus` and it seems to be working for me.
Do you have both the latest version of `find-and-replace` and `autocomplete-plus`?
Does autocomplete not work while searching or not at all?

nathansobo:
Hey. Sorry for the confusion. Thought we had upgraded `autocomplete-plus` but it was reverted due to a test failure in `autocomplete-snippets`. When I linked the latest it worked. Working on upgrading the bundled `autocomplete-plus` now. Thanks again for this!

nathansobo:
@wvanlint After using this a bit, I'm finding the behavior of confirming autocomplete on `enter` to be really problematic in the find and replace editor. Hitting `enter` is used to start searching and so autocomplete ends up getting in the way of running searches.

wvanlint:
@nathansobo Hmm, that makes sense. Do you have any suggestions?

To me, it felt similar to the behavior in the text editor itself, where you need to resolve the autocompletion before you can insert a newline. There are also options for `autocomplete-plus` that people may already use such as `enableAutoActivation` and `confirmCompletion`.

nathansobo:
If we could find a way to make the specificity of the find and replace confirm binding higher than autocomplete's I think that would be really good.

nathansobo:
@wvanlint Hey, I'm really sorry, but I've reverted this pull request for now until we can work out some usability issues related to the shadowing of `enter` as the confirmation binding for find and replace.

I tried to add a higher-specificity selector for the find and replace editor that would toggle `core:confirm` instead of `autocomplete-plus:confirm`, but even when I do this, the autocomplete remains open after confirming a search, which is weird. We need it to close once the find proceeds, and I'm not sure how we're going to achieve that.

Until we solve the usability issue with `enter` I would happily merge a new version of this PR that adds an option for autocomplete in the find editor that defaults to off. But unfortunately I'm too swamped with other stuff to add this myself.

Thanks for the contribution. I do think we'll figure out how to make this a good experience but it's not good enough yet to make it the default today.

wvanlint:
@nathansobo Hey, sorry for the delay!
I added an option for now in #947.

I do still wonder about making the `enter` from `find-and-replace` a higher priority over `autocomplete-plus` while for other keybindings such as `up` and `down` `autocomplete-plus` has the priority over `find-and-replace`.

nathansobo:
Yeah, you're right that the problem may extend beyond `enter`, though I think `enter` has been the biggest show-stopper in my personal usage. I think up/down isn't as bad because at least personally, I usually access the history when the find editor is empty, so there's no conflict with the menu.

I think to get the option on by default, we need enter to both confirm search and dismiss the autocomplete menu. It might be possible to dismiss the menu on `core:confirm` to achieve this, but we'll need to play with it. I'm open to exploring it.

PR #927, opened 2017-07-12
Context lines styling

simurai:
### Description of the Change

This styles the context lines a bit more:

Before | After
--- | ---
screen shot 2017-07-12 at 10 55 54 am | screen shot 2017-07-12 at 10 55 03 am

### Alternate Designs

It would be nice if the context lines and line numbers would also have the same font family as the `.match-line`. It could be done by inlining the font-family to a parent element, maybe `ol.matches`? But I couldn't figure out how. 😞  

### Benefits

It should make it easier to see where new matches start.

### Possible Drawbacks

Context lines are now more subtle and might not have enough contrast. Also, depends on the theme used.

### Applicable Issues

Continues from #908

/cc @dirk-thomas

### DIFF
--- a/styles/find-and-replace.less
+++ b/styles/find-and-replace.less
@@ -294,7 +294,8 @@ atom-workspace.find-visible {
     }
 
     .search-result {
-      border-top: 1px solid @base-border-color;
+      // box-shadow over a border is used to not affect height calculation
+      box-shadow: inset 0 1px 0 @base-border-color;
     }
 
     .preview {
--- a/styles/find-and-replace.less
+++ b/styles/find-and-replace.less
@@ -301,7 +301,16 @@ atom-workspace.find-visible {
       word-break: break-all;
       font-family: @find-and-replace-font;
       white-space: pre;
+      color: @text-color-subtle;
     }
+
+    .match-line .preview {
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The separator line was part of one of my earlier PRs. It was removed from the patch since the added height will need to be considered in the height computation. Even if the tests pass the height computation might be off if it doesn't consider the extra height and especially for long result list that will likely result in problems.

simurai:
> It was removed from the patch since the added height will need to be considered in the height computation.

Good to know. 🤔  Hmm.. maybe we could use an inset `box-shadow` instead of a border? That should be invisible to height calculation. Replaced it.

dirk-thomas:
I tried out the current patch:

context-lines_box-shadow

Is it intentional that each match has `box-shadow` only at the top? I am not sure what would look better - maybe only show the `box-shadow` between matches within the same result? Or also add a `box-shadow` after the last match within a result?

simurai:
> Or also add a `box-shadow` after the last match within a result?

Yeah, that would be better. I added one also after the last match:

image

dirk-thomas:
LGTM. Thank you for improving this.

PR #932, opened 2017-07-22
Added "Open in New Tab" and "Open in New Window" right-click context menu options

tbrockman:
### Requirements

### Description of the Change

Added two context-menu options to right clicking on the paths of search results, "Open in New Tab" ~~and "Open in New Window".~~

image

Can also open new tabs on specific line matches, and already opened tabs will becomes focused after subsequent open attempts:

image

~~Note: Open in New Window not implemented for specific line matches as this does not appear to be easily accomplished within the current API.~~

Includes tests to guarantee that the context menu commands call the
correct functions with the intended parameters, as well as ensure that
right-clicks have no unexpected consequences with the new functionality (such as
expanding and contracting of selected results). But it should be noted that I'm still familiarizing myself with the API and tests, so it's possible that the tests may not be entirely accurate due to that lack of familiarity.

### Alternate Designs

None.

### Benefits

Ability to open multiple results from directory searches without being required to navigate back to search results view. Allows for some nice ~~window and~~ tab organization and facilitates workflows that involve opening and editing multiple files matching certain search criteria.

### Possible Drawbacks

People who frequently use "Copy" and "Copy Path" might accidentally click on the newly added menu options.

### Applicable Issues

N/A. This was just a feature I'd wanted myself in the past, ideally others find it useful.

### DIFF
--- a/lib/project/results-view.js
+++ b/lib/project/results-view.js
@@ -90,8 +90,7 @@ class ResultsView {
         'core:confirm': this.confirmResult.bind(this),
         'core:copy': this.copyResult.bind(this),
         'find-and-replace:copy-path': this.copyPath.bind(this),
-        'find-and-replace:open-in-tab': this.openInTab.bind(this),
-        'find-and-replace:open-in-window': this.openInWindow.bind(this)
+        'find-and-replace:open-in-new-tab': this.openInNewTab.bind(this),
       })
     );
   }
@@ -529,7 +528,7 @@ class ResultsView {
     atom.clipboard.write(relativePath);
   }
 
-  async openInTab() {
+  async openInNewTab() {
     if (this.selectedRowIndex !== -1) {
       const result = this.selectedRow();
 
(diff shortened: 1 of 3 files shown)


50Wliu:
I think for this feature to be useful for me it'd need to have the context menu show up on individual results as well.

> Fixes might include forcing the tab into focus when a new tab is selected

I think this makes the most sense.

tbrockman:
That's fair, I'll work that into the pull request.

Now opening new tabs to a specific line is already supported by the Atom API, but do you know if there exists any sort of way to do the same with a new Atom window?

tbrockman:
Came back to update the code I'd written, fixed merge conflicts, updated calls to no longer existing functions, and then squashed everything into one neat commit.

If there's anything that needs to be changed or added in terms of tests or code, let me know.

lee-dohm:
Thanks for the help! Apologies for taking so long to get back to you.

* Please remove Open in New Window since it isn't completely implemented
* Please rename command to `find-and-replace:open-in-new-tab`

Thanks again for the contribution!

tbrockman:
No worries, glad to hear back!

Done and done, let me know if there's anything else.

Issue #133, opened 2014-02-28
Search in search results. Searcheption.


hkdobrev:
Discourse thread: http://discuss.atom.io/t/searchception-missing/1779

It expands on #67.

In Sublime when you search in a project/folder you get the search results in a normal buffer. You could perform any text manipulation in there and also search. Atom should do that as well.

park9140:
:+1:  This would be great.

townie:
This is a must have feature!

alimoosavi15:
This is the only feature I'm missing from Komodo IDE.

rahlzel:
It's sad to find this request started back in 2014 with nothing to show for it now in 2016. We do need this functionality. As @hkdobrev mentioned, it's one of the big reasons I use Sublime. With massive projects, sometimes we have a vague clue to find the code we're looking for, but that clue gives us a few hundred results that we then need to dig through. Being able to search the search results would help us narrow things down tremendously.

Thanks!

eliveikis:
Same here, the one feature that keeps me using Sublime.

sebfindling:
Any news on this?? @damieng @travco @lee-dohm @codingbelief

damieng:
I think what Ben said still stands.  It's not on the core team's roadmap so it would need to be a community PR to be considered.

travco:
@sebfindling  Expanding on Ben's earlier answer (on the technical side):

> Project search results are not text in atom. ...

> ... the current FindModel only works with pane items that have a buffer property.

As far as I know, to make this viable, one of two things needs to be done. Either adapt entire find model (as it is) to make it able to read in non-buffer content (and figure out how to differentiate what we can and can't handle searchwise). Or change the output of search results to be a buffer - and still be as useful as it was (not sacrificing/backpedaling on features). 

Unless there's a trivial third option, I'm personally not comfortable with atom's internals (or this plugin's for that matter) to dive into something like this with any degree of confidence myself.

osscar:
bump

GregariousJB:
I found a solution. I switched to [VSCode](https://code.visualstudio.com/).

In all seriousness, being able to open the search results in a text window would solve this, like VSCode and Sublime.

Here's an example I whipped up using VSCode:

![image](https://i.imgur.com/cCSXNfH.png)

osscar:
Exactly that... but I don't wanna change

PR #933, opened 2017-07-23
Search results as text

jchavarri:
Hi! 👋 I have started using Atom lately, and I love it :) but I was missing this feature so much, so taking a stab at building something like it here.

### Description of the Change

Adds a new class `ResultsTextViewManager` that uses `TextEditor` in order to show the results as text.

It is _heavily_ influenced by Sublime Text search results. At this point, it features:

- Shows the results as text, showing file paths, and line numbers for each result
- Merges results that are on the same line for a more continuous view
- Configurable number of lines before and after result via the already existing `searchContextLineCountBefore` and `searchContextLineCountAfter`
- Double click on a result line to open the result file at that specific line
- Un-intrusive, "opt-in" configuration via a newly added `find-and-replace.findResultsAsText` config key
- A grammar created "on the fly" to highlight file paths and line numbers
- The search results pane gets the right mag icon + title
- Uses `shouldPromptToSave` to not prompt the user when closing the search results pane

screen shot 2017-07-24 at 00 09 57

### Alternate Designs

_Disclaimer: I have no experience developing Atom plugins or with the Atom codebase at all_ 😁 _so there is probably plenty of room for improvement._

I tried to not touch the existing `ResultsModel`, and just add any needed modifications on the text output at the "rendering" step (i.e. just before setting the text on the `TextEditor` instance). I also tried to not subclass `TextEditor` as there seems to be a lot of logic relying on that class that might fail when subclassing.

### Benefits

The main benefit of implementing the results as text with an "opt in" approach is that users that are used to the existing web based results pane won't see any changes.

The benefits of having results as text are many:
- Use buffer search to search on the results (aka "Searchception")
- Copy text from the results
- Save the results in a text file

### Possible Drawbacks

I don't see any 🤔 I was expecting to see performance issues but it seems to be as performant as the web based search results.

### Applicable Issues

https://github.com/atom/find-and-replace/issues/133 _Search in search results. Searcheption._

Also related: https://discuss.atom.io/t/searchception-missing/1779/3.

### Todo

- Add more tests
- Add a separate set of markers for the project results to keep highlighting them in case a buffer search starts (right now, the buffer search takes over any existing markers)
- “Progressive” results that appear while searching (i.e. use `onDidSearchPaths`)
- Add other listeners: `onDidClear`, `onDidClearReplacementState`, `onDidErrorForPath`... it's not clear to me what they do yet.
- Hide gutter lines: I wasn't able to override the user setting
- Open files with `pending` on double click
- Use `projectSearchResultsPaneSplitDirection` when opening files on double click (not sure what this config does).
- Fix a bug where the grammar won't be applied correctly when searching for a second time with the same query string as the first time, after closing the search results pane
- Fix a bug where when reopening Atom, an orphan `untitled` pane opens on launch if the `Project Find Results` pane was being shown before closing.

### Status / questions

While this is clearly not ready to merge, I would really appreciate a first review or some pointers from the maintainers about the general strategy. Some other questions I have, and couldn't find answers are:

- ES6 vs Coffeescript: I am assuming that ES6 is the right choice due to the most recent files using it, is that correct?
- How does one enable ES6 linting in Atom when working with the `find-and-replace` project? There doesn't seem to be any `.eslintrc` file or `package.json` config.

Thanks!


50Wliu:
_Disclaimer: Any comments I make do not mean that I will be the one reviewing and merging this PR.  Just trying to be helpful._

* ES6 is correct.  A gradual migration is in progress from Coffeescript to JavaScript.
* I'm not sure on linting.  Some projects are using a custom ESLint config, some are using Standard, and some are using eslint-config-fbjs-opensource.  I don't think a consensus has been reached yet, but Standard seems like a safe bet in the interim.

PR #938, opened 2017-08-10
Add preserve case option for buffer replace. References #165.

paradoxxxzero:
### Description of the Change

This PR adds a `preserveCaseOnReplace` option (off by default) which keeps the replaced text case. For exemple, when replacing `test` with `case`:
`test Test TEST` will be replaced with: `case Case CASE`.

This is made by swapping the replacement right before putting it in the buffer in `buffer-search.js`. The `preserveCase` algorithm is trivial and defined in `project/util.coffee`.

**This only works in buffer search. The project replace is done in atom core and should be addressed in a separate ticket.**

### Alternate Designs

One could have added a button for this option, but a plugin configuration was chosen since it concerns only replacing (not searching) and it might not be toggled very often.

### Benefits

Huge benefits arise when replacing variables name that are in different case, for instance replacing in one pass `client`, `getClient` and `MAX_CLIENT` with `person`, `getPerson`, `MAX_PERSON`.

### Possible Drawbacks

This adds a boolean check for each replace, low performance impact though.

### Applicable Issues
#165

### DIFF
--- a/lib/buffer-search.js
+++ b/lib/buffer-search.js
@@ -1,6 +1,7 @@
 const { Point, Range, Emitter, CompositeDisposable, TextBuffer } = require('atom');
 const FindOptions = require('./find-options');
 const escapeHelper = require('./escape-helper');
+const Util = require('./project/util');
 
 const ResultsMarkerLayersByEditor = new WeakMap;
 
@@ -92,6 +93,7 @@ class BufferSearch {
     if (!markers || markers.length === 0) return;
 
     this.findOptions.set({replacePattern});
+    const preserveCaseOnReplace = atom.config.get('find-and-replace.preserveCaseOnReplace')
 
     this.editor.transact(() => {
       let findRegex = null
@@ -104,9 +106,10 @@ class BufferSearch {
       for (let i = 0, n = markers.length; i < n; i++) {
         const marker = markers[i]
         const bufferRange = marker.getBufferRange();
-        const replacementText = findRegex ?
+        let replacementText = findRegex ?
           this.editor.getTextInBufferRange(bufferRange).replace(findRegex, replacePattern) :
           replacePattern;
+        replacementText = preserveCaseOnReplace ? Util.preserveCase(replacementText, this.editor.getTextInBufferRange(bufferRange)): replacementText
         this.editor.setTextInBufferRange(bufferRange, replacementText);
 
         marker.destroy();
--- a/lib/buffer-search.js
+++ b/lib/buffer-search.js
@@ -106,10 +106,9 @@ class BufferSearch {
       for (let i = 0, n = markers.length; i < n; i++) {
         const marker = markers[i]
         const bufferRange = marker.getBufferRange();
-        let replacementText = findRegex ?
-          this.editor.getTextInBufferRange(bufferRange).replace(findRegex, replacePattern) :
-          replacePattern;
-        replacementText = preserveCaseOnReplace ? Util.preserveCase(replacementText, this.editor.getTextInBufferRange(bufferRange)): replacementText
+        const replacedText = this.editor.getTextInBufferRange(bufferRange)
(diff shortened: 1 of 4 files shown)


lib/buffer-search.js
```diff
@@ -104,9 +106,9 @@ class BufferSearch {
       for (let i = 0, n = markers.length; i < n; i++) {
         const marker = markers[i]
         const bufferRange = marker.getBufferRange();
-        const replacementText = findRegex ?
-          this.editor.getTextInBufferRange(bufferRange).replace(findRegex, replacePattern) :
-          replacePattern;
+        const replacedText = this.editor.getTextInBufferRange(bufferRange)
+        let replacementText = findRegex ? replacedText.replace(findRegex, replacePattern) : replacePattern;
+        replacementText = preserveCaseOnReplace ? Util.preserveCase(replacementText, replacedText): replacementText
```
bmperrea: Nice

lib/project/util.coffee
```diff
@@ -36,7 +36,21 @@ showIf = (condition) ->
...
+preserveCase = (text, reference) ->
+  # If replaced text is capitalized (strict), capitalize replacement
+  if reference is capitalize(reference.toLowerCase())
+    capitalize(text)
+
+  # If replaced text is uppercase, uppercase replacement
+  else if reference is reference.toUpperCase()
+    text.toUpperCase()
+  else
+    text
```
bmperrea: This seems like a difficult feature to design. You have implemented preservation of All CAPS, and of Proper case. I would have expected lower case to be preserved as well. It might be worth thinking through a few more cases, such as strings with chars other than spaces and letters (e.g. one could imaging replacing snake_case and SNAKE_CASE at the same time, one being recognizably lower case and the other upper). camelCase might be interesting too if both the find and replace had that form, although maybe now I'm getting too fancy. I would prefer to open this discussion a little bit, though, such as with an issue for this feature.

spec/buffer-search-spec.js
```diff
@@ -514,3 +514,191 @@ describe("BufferSearch", () => {
...
+        CCC DDD FoO
+        -----------
+        FOO Bbb cCc
+        Ddd FoO Bbb
+        ccc DDD FoO
+        -----------
+      `);
+    });
+  });
+});
```
bmperrea: well done

lib/project/util.coffee
bmperrea: You have support for 

- Title Case
- UPPER CASE

I suggest you add

- lower case
- Sentence case

These are the main ones in [Wikipedia/Letter_case](https://en.wikipedia.org/wiki/Letter_case), and ones that can be recovered from lower case. Then the user can set the default (what they put in the replace box) to whatever mixed case they want otherwise. I realize not supporting "Sentence case" was by design, but I would expect it to keep that case. Thanks!

paradoxxxzero:
First, thanks for the code review!
Ok so now when replacing `aAa bBb` by `xxX yYy`, it goes like this:
```rust
aaa bbb -> xxx yyy  # lower case
Aaa bbb -> Xxx yyy  # Sentence case
aaa Bbb -> xxX yYy  # ? -> default
Aaa Bbb -> Xxx Yyy  # Title Case
AAA BBB -> XXX YYY  # UPPER CASE
aaA bbb -> xxX yYy  # ? -> default
aaa bbB -> xxX yYy  # ? -> default
aaA bbB -> xxX yYy  # ? -> default
```
Is that what you wanted?

paradoxxxzero:
Is there anything else that I could do to make it hit a release?

Le-Stagiaire:
Aaaah that's what I was looking for ! Hope to see it in the next release.
Shouldn't it be merge already though ?

jedwards1211:
I made a PR that supports a wider range of cases like snake case, camel case, and kebab case.  Could we close this in favor of it?

darangi:
Thanks @paradoxxxzero

PR #939, opened 2017-08-10
Use multiline flag when replacing

### LINKED ISSUE #622: Nothing replaced when regex contains ^ or $
- I right-click on a folder and chose "Search in directory"
- I wrote my search phrase `^###` (regex)
- I wrote my replace phrase `RR2`
- I clicked find
- `129 results found in 29 files for ^###`
- I clicked replace all
- `This will replace '^### ' with 'RR2' 129 times in 29 files`
- I clicked OK
- Searching ...
- `Nothing replaced`

Indeed, none of my files were modified.

What could lead to such a behavior?

I just checked without the regex option, and it replaces the text correctly (but I need regex here).

50Wliu:
### Requirements

### Description of the Change

The multiline flag changes `^` and `$` to line-based rather than file-based.  

### Alternate Designs

None.

### Benefits

Replacing with `^` and `$` will work as expected.

### Possible Drawbacks

None.

### Applicable Issues

Fixes #622
Refs atom/atom#15224

I believe the build will fail since atom/atom#15224 is only on Atom 1.21.0.

### DIFF
--- a/spec/project-find-view-spec.js
+++ b/spec/project-find-view-spec.js
@@ -1369,6 +1369,11 @@ describe('ProjectFindView', () => {
 
       describe("when the find field contains a ^ or a $ and the regex option is enabled", () => {
         it("correctly replaces all matches", async () => {
+          // TODO: Remove version check when Atom 1.21 reaches stable
+          if (parseFloat(atom.getVersion()) < 1.21) {
+            return;
+          }
+
           atom.commands.dispatch(projectFindView.element, 'project-find:toggle-regex-option');
           projectFindView.findEditor.setText(';$');
           atom.commands.dispatch(projectFindView.element, 'core:confirm');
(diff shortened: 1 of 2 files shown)


spec/project-find-view-spec.js
```diff
@@ -1369,6 +1369,11 @@ describe('ProjectFindView', () => {
 
       describe("when the find field contains a ^ or a $ and the regex option is enabled", () => {
         it("correctly replaces all matches", async () => {
+          // TODO: Remove version check when Atom 1.21 reaches stable
+          if (parseFloat(atom.getVersion()) < 1.21) {
```
C-Bouthoorn: I'm not sure if they exist, but Atom 1.3, 1.4, 1.5 all the way to 1.9 would fail this check.

50Wliu: I'm aware.  Atom 1.9 was released approximately one year ago and is not supported.

Issue #417, opened 2015-06-27
feature request: allow basic javascript in find and replace function


Billybobbonnet:
Hello,

My issue is quite simple, I have a list of countries in  ISO 3166-1 alpha-2 format. here is how it looks:

``` json
{
  "name": "Afghanistan",
  "code": "AF",
  "country-code": "004"
}, {
  "name": "Åland Islands",
  "code": "AX",
  "country-code": "248"
}, {
  "name": "Albania",
  "code": "AL",
  "country-code": "008"
},
```

I want to change all the codes to their lowercase equivalent. I can select them using a simple regex find string `"[A-Z]{2}"` 

I would love a way to apply a `matchedString.toLowerCase()` on every string matched by my regex. Or maybe something like `this.toLowerCase();`

Thanks!

dorian-marchal:
You can use <kbd>Ctrl+K U</kbd> to change the selection to uppercase and <kbd>Ctrl+K L</kbd> to lowercase.

Billybobbonnet:
I just did a click on `find` + `Ctrl+K L ` 294 times. It was surely shorter than doing it "manually". But still, I fell ashamed :wink: 

I love atom and I want to stick with it. I must say that this is the kind of feature that I would like to have in my main development tool.

Keep up the good work!

davelnewton:
A partial solution would be to allow Sublime-style replacement expressions, e.g., https://manual.macromates.com/en/regular_expressions#replacement_string_syntax_format_strings. That makes easy stuff super-easy. It's one of the few things I really miss about Sublime. That and keyboard macros :(

benogle:
> A partial solution would be to allow Sublime-style replacement expressions

Do you mean replacing `ab([c]+)` with `de$1` style of replacement? Atom currently supports these. Or is there something in that link that I am missing? A solution to this issue without allowing js would be interesting.

davelnewton:
No, I mean to allow Sublime-style modifiers like `\u` to upper-case as mentioned in the "case foldings" section:

> It is possible to convert the next character to upper or lowercase by prepending it with \u or \l. This is mainly useful when the next character stems from a capture register. Example:

    Find: (<a.*?>)(.*?)(</a>)
    Replace: $1\u$2$3

> You can also convert a longer sequence to upper or lowercase by using \U or \L and then \E to disable the case folding again. Example:

    Find: (<a.*?>)(.*?)(</a>)
    Replace: $1\U$2\E$3

[...]

PR #944, opened 2017-08-22
Feature: Javascript Function mode for replace

bmperrea:
### Requirements

### Description of the Change

This PR exposes the ability to pass a function as the second argument to javascript's String.prototype.replace() function. This can be useful if you want to replace snake_case with camelCase, for instance with `text.replace(/_([a-z])/g, (m, m1)=>m1.toUpperCase())`. The design is such that one types a javascript function into the Replace text box. 

Careful code design was taken to catch all possible errors, without slowing down the current implementation, and also making it efficient in the new function mode. Careful error analysis catches errors on function compilation and function evaluation, alerting the user appropriately in the error box.

### Alternate Designs

1. For the overall feature design, javascript functions were chosen for scriptable replacements. Alternate feature designs were suggested in #417 such as using a regex-style replace object such as with \U and \L for uppercase and lowercase (as in Sublime), and maybe match syntax `\0` and `\1`. One reason we decided to stick with the original suggestion was that it is immediately supported in javascript, the language in which atom is currently implemented. Specifically, the replace method currently relies on Javascript's string.prototype.replace method. Another is that the function syntax allows more functionality with full freedom to write one's own replace function. The biggest drawback is that one needs to use Javascript, which may not be familiar to all. I would point out that regex replace syntax is not completely standardized and would also be often unfamiliar.

2. We added a button on the replace line to toggle function mode. One could consider adding this to the find options buttons, or otherwise.

3. String.prototype.replace always casts it's result as a string. We chose to override this by first checking that the result is actually a string to help avoid errors such as returning an undefined key, or variable. One can still manually cast to a string in the user-written replace function, as in the error handling demo below.

### Benefits

One can programmatically choose the replace text based on regex matches (or a single text match). Use cases include changing case and using simple logic to chose the replacement text.

### Possible Drawbacks

This makes for more UI (an extra button - see demos below). It also makes the buffer_search code more involved.

### Applicable Issues

This solves #417.

### DIFF
--- a/keymaps/find-and-replace.cson
+++ b/keymaps/find-and-replace.cson
@@ -38,6 +38,7 @@
   'cmd-alt-c': 'find-and-replace:toggle-case-option'
   'cmd-alt-s': 'find-and-replace:toggle-selection-option'
   'cmd-alt-w': 'find-and-replace:toggle-whole-word-option'
+  'cmd-alt-f': 'find-and-replace:toggle-function-option'
 
 '.platform-win32 .find-and-replace, .platform-linux .find-and-replace':
   'shift-enter': 'find-and-replace:show-previous'
@@ -51,12 +52,13 @@
   'cmd-alt-/': 'project-find:toggle-regex-option'
   'cmd-alt-c': 'project-find:toggle-case-option'
   'cmd-alt-w': 'project-find:toggle-whole-word-option'
+  'cmd-alt-f': 'find-and-replace:toggle-function-option'
 
 '.platform-win32 .project-find, .platform-linux .project-find':
   'ctrl-enter': 'project-find:confirm'
   'ctrl-alt-/': 'project-find:toggle-regex-option'
   'ctrl-shift-c': 'find-and-replace:toggle-case-option'
-  
+
 '.find-and-replace, .project-find, .project-find .results-view':
   'tab': 'find-and-replace:focus-next'
   'shift-tab': 'find-and-replace:focus-previous'
--- a/lib/buffer-search.js
+++ b/lib/buffer-search.js
@@ -94,21 +94,51 @@ class BufferSearch {
 
     this.findOptions.set({replacePattern});
     var useFunction = this.findOptions.useFunction;
+    var replaceMap;
 
     if(useFunction){
+      var err;
       var replaceFcn;
       try {
         replaceFcn = vm.runInThisContext(replacePattern);
-      } catch(err) {
+      } catch(error) {
+        err = error;
         err.message = "Function didn't compile: " + err.message;
         return this.emitter.emit('did-error', err);
       }
       if(replaceFcn && typeof replaceFcn === "function") {
+        let findRegex = null
+        if (this.findOptions.useRegex) {
+          findRegex = this.getFindPatternRegex();
+          if (!useFunction){
+            replacePattern = escapeHelper.unescapeEscapeSequence(replacePattern);
+          }
+        }
         replacePattern = replaceFcn;
+        replaceMap = markers.map(marker=>{
+          const bufferRange = marker.getBufferRange();
+          const currentText = this.editor.getTextInBufferRange(bufferRange);
+          let replacementText;
+          try {
+            replacementText = currentText.replace(findRegex || currentText,
+              replacePattern);
+          } catch(error) {
+            err = error
+          }
+          if(!replacementText){
+            err = {
+              message: "Function returned " + (replacementText
+                && replacementText.toString ? replacementText.toString() : replacementText)
+            };
+          }
+          return { marker, bufferRange, replacementText};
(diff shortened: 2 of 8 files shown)


bmperrea:
Other things that would be nice (but can probably wait for another PR):

1. Extend to project replace
2. Make the icon height match the input box and buttons better
3. Colorize the replace function box red when there is a function-related error - this should be done in the themes repos: [light](https://github.com/atom/one-light-ui/pull/86/files) and [dark](https://github.com/atom/one-dark-ui/pull/183), in a way similar to the PRs mention in [this issue](https://github.com/atom/one-dark-ui/issues/150)].

bmperrea:
camel_to_snake

bmperrea:
Demo of error handling
function_error_demo

bmperrea:
@50Wliu - can you change the tag to "needs-review" - thanks!

monster860:
This would be a really great feature. Why is it not been touched forever?

curiousercreative:
Would love to see this merged in :)

jedwards1211:
Nudge

lee-dohm:
Thank you for your contribution.

We've reviewed this and while we can see that there are definite benefits for certain workflows, this is not functionality that we want to maintain long-term. Because of this, we're going to close this PR and the associated issue as well. Our apologies for taking so long to come to a decision on it.

Thanks for your contribution and your interest in Atom.

RedHatter:
This should be released as a separate plugin.

RedHatter:
I have released a fork of find-and-replace with this PR applied.

https://atom.io/packages/fn-find-and-replace

robfrawley:
The contribution implemented by @bmperrea in this PR is critically useful, and while it differs from Sublime Text's handling, I find no fault in this alternate syntax. In my opinion, it represents an excellent feature addition to Atom, one present in countless other editors from Notepad++ to Sublime Text and many more.

> We've reviewed this concept at length and [...] this isn't something that the team wants to maintain long-term[, though] we do believe that this would make an exciting very interesting community package... -@lee-dohm

I'm incredibly disappointed that @lee-dohm closed this PR, and additionally, I'm aggrieved by the complete lack of official discussion by Atom maintainers regarding this feature. The closing comment by @lee-dohm is so generic it could *literally* be used to terminate any PR. No official comment provides any reasonable context or any useful explanation of their final decision. It does not explain why or how they arrived at their position, and for those submitting new PRs with similar or related functionality, it offers no guidance whatsoever.

Despite @lee-dohm's assertions, I fail to fathom how this minor feature could create any measurable code maintenance overhead that negatively drives down productivity on other project aspects.

__I cannot overstate the importance of this feature: it is critical to many workflows__. The ability to control the case of search replacement matches is available in countless editors from Sublime Text to Notepad++ and many more. I find it particularly significant that a *tiny community-driven editor like Notepad++ has the resources to create and maintain this feature* while Atom, with the corporate backing of GitHub, does not. The given explanation is laughable on its face.

My prior employer, Scribe Inc, directly illustrates the corporate importance this feature represents (either in this implementation or a different one). Scribe specializes in XML-based publishing using a custom [SCML](https://github.com/scribenet/scml_dtds) DTD (*SC*ribe x*ML*). This workflow enables automatic transformations from the ScML source to any of the many intermediate and final publishing formats required (such as InDesign for print books, HTML for e-books, HTML for online web release, Word documents for authors/editors, etc.). 

The source material received from authors and their publishers is never correctly tagged or formatted to the required specification. Therefore, the initial ScML version requires extensive manual intervention. Creating the first DTD-compliant version of the material is a difficult task that relies on the case-control-aware-replace functionality of Sublime Text. Moreover, relying on plugins is not generally allowed because the initial ScML-creation step is completed by different individuals depending on the client. These procedures are performed either by the author, their publisher, an internal employee, or even an external contractor. __Without case-control search and replace, *Scribe would be unable to switch their ScML editor to Atom in their workflow*, regardless of their preference.__

I sincerely hope you further continue this discussion and that this feature gets included within the core Atom codebase.

*Note: Aside from this feature itself, I noticed something else troubling. Why are *all four comments* supporting this functionality __marked as *spam comments* that are hidden by default__? See https://github.com/atom/find-and-replace/pull/944#issuecomment-386622409, https://github.com/atom/find-and-replace/pull/944#issuecomment-398176491, https://github.com/atom/find-and-replace/pull/944#issuecomment-412891405, and https://github.com/atom/find-and-replace/pull/944#issuecomment-444010792. They all advocate for the inclusion of this feature in Atom and are not off-topic or otherwise spam messages. It seems suspicious that they have been hidden by default, effectively limiting the exposure of positive responses to this feature. Perhaps my assumption is presumptuous or even pessimistic. Regardless, I do trust this error is quickly remedied irrespective of the point of view taken by the comment.*

PR #947, opened 2017-08-26
Option to enable autocompletion

wvanlint:
Continuation of #922

### DIFF
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -122,18 +122,24 @@ module.exports =
     new Disposable ->
       FileIcons.resetService()
 
-  requestAutocompletions: ->
-    disposable = @autocompleteWatchEditor?(@findView.findEditor, ['default'])
-    if disposable?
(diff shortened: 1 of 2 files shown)


50Wliu:
Here's something I noticed while this was enabled:
Searching for something that triggers autocomplete, then confirming the search without also confirming the autocomplete (e.g. by using the Find button) does not remove the autocomplete list.  That was visually jarring for me as I expected that the autocompletions would disappear after starting the search.

nathansobo:
@50Wliu do you think it's a big enough problem that we should hold this back? I think I'm okay letting people play with it as the non-default, but I agree that it needs more work to be fully usable.

50Wliu:
I think that should be fixed before the option becomes on by default, but I'm fine with dealing with it while it's still experimental.

nathansobo:
Thanks for adding the option @wvanlint. Any interest in taking on some of the usability challenges so we can enable it by default?

PR #954, opened 2017-09-01
Fix error when selecting result after changing its corresponding buffer

### LINKED ISSUE #953: 1.20 Beta: Editing contents of results file breaks navigation.
### Prerequisites

### Description
In the current beta version of Atom, navigating to a find-in-project result and changing the contents of the file causes the package to throw an error (`TypeError: Cannot read property '0' of undefined`) on subsequent attempts to jump to search results. Here's a quick screencast, with the v1.20 beta in safe mode, demonstrating what I mean

screenflow

### Steps to Reproduce

1. Perform a project-wide search with the find-and-replace panel.
2. Click a hit in the "Project Find Results" view.
3. Edit and save the file.
3. Click another hit in the "Project Find Results" view.
4. No bueno.

**Expected behavior:** Atom should jump to the second hit without throwing an error.

**Actual behavior:** Error pops up and we're stuck where we are.

**Reproduces how often:** 100%

### Versions
```
~ ❯❯❯ atom-beta --version
Atom    : 1.20.0-beta5
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0

~ ❯❯❯ apm-beta --version
apm  1.18.4
npm  3.10.10
node 6.9.5 x64
python 2.7.10
git 2.14.1

~ ❯❯❯ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.12.6
BuildVersion:	16G29
```

### Additional Information
💋

as-cii:
Fixes #953

With this pull-request we are normalizing the match result ranges to always be `Range` objects. Previously, ranges were being expressed sometimes as arrays and sometimes as `Range` objects.

@maxbrunsfeld: I wonder if this has anything to do with the recent changes to the buffer APIs. I think it's okay to paper over the problem in this specific package, but I worry other packages may break for similar reasons.

/cc: @iolsen @ungb @rsese @Ben3eeE


Ben3eeE:
This does fix the issue :+1:

rsese:
Also confirmed fix on macOS 10.12.6 using https://github.com/atom/atom/releases/tag/untagged-e8e64b7f05f18088017b 💻

as-cii:
Okay, I think I won't cherry-pick this one onto stable and beta, mainly because there are other changes in master that I don't want to include in a hotfix (e.g. coffee to js conversion, find-and-replace autocompletion, etc.).

I will still go ahead and merge it once it goes green, so that it can bake a little bit on master and then just go out on beta when rolling the railcars.

PR #959, opened 2017-09-14
Fix project-find:toggle-case-option keymap on Windows & Linux

segevfiner:
### Requirements

### Description of the Change

The keymap for `project-find:toggle-case-option` was wrong and pointed to the `find-and-replace:toggle-case-option` instead.

### Benefits

Bug fix. The keymap will now work.


50Wliu:
:+1:

PR #968, opened 2017-11-01
Fixed #950

### LINKED ISSUE #950: Disabling and re-enabling the package makes select-next command select two next occurences of word
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:

### Description

As reported in https://github.com/mehcode/atom-project-plus/issues/98, after programatically disabling and then re-enabling this package, the `find-and-replace:select-next` command will select the next two occurences of the highlighted word. Repeating the disable/re-enable procedure will select the next three occurences. Do it again and it will select the next 4, and so on.

The reason that [atom-project-plus](https://github.com/mehcode/atom-project-plus) disables/re-enables this package is that it otherwise, find-and-replace won't be able to search any file in the newly opened project.

I don't know which one of these issues is easier to fix, but my bet is on the former.

### Steps to Reproduce

1. Install atom-project-plus
2. Open a project normally in Atom
3. Switch projects using atom-project-plus
4. Highlight a word and trigger the `find-and-replace:select-next` command

**Expected behavior:**
The next occurence of the word should be selected

**Actual behavior:**
The two next occurences of that word is selected

**Reproduces how often:**
Every time

### Versions
```
Atom    : 1.19.4
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0

find-and-replace: 0.208.3
```

### Additional information

Once again, this seems to be triggered by disabling and re-enabling the find-and-replace package. The relevant code can be found [here](https://github.com/mehcode/atom-project-util/blob/master/index.js#L215-L217)

xream:
Fixed #950

### Requirements

### Description of the Change

Commands should be removed when the package is deactivated.

### Benefits

Fixed #950

### DIFF
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -107,7 +107,7 @@ module.exports =
         panelToToggle.show()
         postToggleAction?()
 
-    atom.commands.add '.editor:not(.mini)',
+    @subscriptions.add atom.commands.add '.editor:not(.mini)',
       'find-and-replace:select-next': (event) ->
         selectNextObjectForEditorElement(this).findAndSelectNext()
       'find-and-replace:select-all': (event) ->


50Wliu:
Thanks!

PR #976, opened 2017-11-18
Ensure that word boundaries are matching start/end of query

tribut:
### Description of the Change

When the original expression is not grouped and a regex-query is used,
the boundaries might not be matched. For example:

    this|is|a|test

was matching all letters "a", not just the word "a" - even if "whole
word" is enabled.

screenshot from 2017-11-18 18-51-32
screenshot from 2017-11-18 18-52-00

### DIFF
--- a/lib/find-options.coffee
+++ b/lib/find-options.coffee
@@ -61,6 +61,6 @@ class FindOptions
     else
       expression = _.escapeRegExp(@findPattern)
 
-    expression = "\\b#{expression}\\b" if @wholeWord
+    expression = "\\b(?:#{expression})\\b" if @wholeWord
 
     new RegExp(expression, flags)
--- a/spec/project-find-view-spec.js
+++ b/spec/project-find-view-spec.js
@@ -666,7 +666,7 @@ describe('ProjectFindView', () => {
 
         await searchPromise;
         expect(projectFindView.refs.wholeWordOptionButton).toHaveClass('selected');
-        expect(atom.workspace.scan.mostRecentCall.args[0]).toEqual(/\bwholeword\b/gim);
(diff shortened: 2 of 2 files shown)


lee-dohm:
Thanks for the work you put into this and the contribution.

Unfortunately, regular expressions are simply too complex for us to feel comfortable guaranteeing that "Whole Word" mode will be compatible with all possible valid regular expressions. It would be a huge amount of work to validate and then maintain, especially as we ultimately don't control the underlying regular expression library since it comes to us from Chromium. Because of this, we don't want to accept this functionality.

On the other hand, this does perfectly illustrate a bug that currently exists in the application. The simplest fix for this bug, however, is not to attempt to make "Whole Word" mode work with all possible valid regular expressions, but to make "Whole Word" mode and "Regex" mode mutually exclusive. I've opened [an issue](https://github.com/atom/find-and-replace/issues/1044) to track this feature.

Thanks again for your contribution.

Open a new issue describing making regex and whole word mutually exclusive selections

jedwards1211:
I can understand your concerns about maintenance, but I think you're overestimating how difficult this will be.  I don't see how any other problems could arise after we wrap the regex in a non-capturing group, as long as we check that the input regex is valid by itself.

jedwards1211:
In any case I guess the Atom team would prefer we rely on plugins to provide Industrial-strength search and replace that is on par with some other editors?

jedwards1211:
I am also wondering if Microsoft is trying to sideline atom development in favor of VSCode, since they recently bought GitHub...

jedwards1211:
@lee-dohm how would any features get added to JavaScript `RegExp`s that would break a `\b(?: )\b` wrapper?  I fail to see how this would be a problem.

PR #986, opened 2018-01-05
Fixed overflow from binary search in /lib/project/results-model.js

angwhen:
### Requirements

### Description of the Change

In the binary search below, the original calculation of mid may lead to overflow if the "low" and "high" variables are too large, and their sum reaches above the maximum number value. I found out about this problem in https://research.googleblog.com/2006/06/extra-extra-read-all-about-it-nearly.html .

### Alternate Designs

I also considered "const mid = low + Math.floor((high - low) / 2)", as opposed to "const mid = Math.floor(low + (high - low) / 2)", since they get the same result. I could not find any benefit of either over the other.

### Benefits

This will prevent binary search from failing on very large arrays, where high and low may both be respectively close the the maximum value.

### Possible Drawbacks

I don't think there are drawbacks unless I made coding errors. However, because Javascript has a very high maximum number value (https://stackoverflow.com/questions/307179/what-is-javascripts-highest-integer-value-that-a-number-can-go-to-without-losin) the change may not be necessary for most realistic cases.

### Applicable Issues

I don't think there are many additional application issues.

### DIFF
--- a/lib/project/results-model.js
+++ b/lib/project/results-model.js
@@ -342,7 +342,7 @@ var binaryIndex = function (array, value, comparator) {
   let low = 0
   let high = array.length
   while (low < high) {
-    const mid = Math.floor((low + high) / 2)
+    const mid = Math.floor(low + (high - low) / 2)
     if (comparator(array[mid], value) < 0) {
       low = mid + 1
     } else {


PoignardAzur:
This PR is made obsolete by #1002. Should be closed.

Issue #984, opened 2017-12-20
Search in Directory includes results from open, modified files outside directory


bhj:
### Prerequisites

* [X ] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

Search in Directory includes results from file(s) outside the directory when the file(s) are open and modified (blue dot). This does not occur if the file(s) are open but unmodified.

### Steps to Reproduce

1. Open a file from the project folder
2. Modify the file without saving
3. Run Search in Directory with a search string said file contains, but choose a folder the file is not in

*Edit by @rsese to add expected/actual behavior*

**Expected behavior:** [What you expect to happen]

No results from the file in step 1.

**Actual behavior:** [What actually happens]

Results from the file in step 1.

**Reproduces how often:** [What percentage of the time does it reproduce?]

100%

### Versions

```
Atom    : 1.23.1
Electron: 1.6.15
Chrome  : 56.0.2924.87
Node    : 7.4.0
apm  1.18.11
npm  3.10.10
node 6.9.5 x64
atom 1.23.1
python 2.7.10
git 2.14.3

macOS 10.13.2
```

anderoonies:
i guess it should be clarified what the desired behavior is before moving forward with this
it makes sense to search buffers you have open, but the saved/unsaved distinction seems like it could be (and has been) confusing.
should search check the present working directory + all open buffers?

anderoonies:
this is a result of:
https://github.com/atom/find-and-replace/blob/3867553a839d4affe651d3669ba0f6d20e4b8dfd/lib/project/results-model.js#L35-L39

the subscription doesn't check if the modified item is part of the workspace's project before adding it to the result.

i'll have a PR up for this sometime soon

bhj:
Thanks for looking into this @anderoonies - it resulted in some unexpected "destructive" actions when doing a Replace All and thinking the scope would be limited to the directory.

rsese:
[...]

PR #988, opened 2018-01-10
Don't watch TextEditors outside of the current project in Project Find

anderoonies:
### Description of the Change

`workspace.scan` only searches the active `project`. However, `ResultsModel` was subscribing to all `TextEditor`s for updating the results of a search, including `TextEditors` that don't belong to the `project`.

To reproduce this, create a hierarchy something like

```
root
 |- 1.txt
 nested
  |- 2.txt
  |- 3.txt
```

and open atom with `atom nested`—specifically setting the project path to be `root/nested`. A Project Find will only include buffers in `nested` (2.txt, 3.txt).
Open 1.txt and modify its contents to include the search term. It will appear in the Results View.

This PR adds a simple check to the `ResultsModel` `workspace` subscription to prevent this from happening.

### Alternate Designs

The subscriptions could be exclusively created with TextEditors that exist in the current project, but managing creating/disposing of those adds complexity and room for error.

### Benefits

Removes confusion/unintuitive behavior around finding and replacing.

### Possible Drawbacks

May have become an anti-pattern, removal of which could cause confusion.
Specs became more complex because of the introduction of a level of nesting.

### Applicable Issues

#984


anderoonies:
@bhj this covers that case, too! good note, thanks for catching it

PR #992, opened 2018-01-17
Fix last result not clickable

### LINKED ISSUE #943: Last result in Project Find Results is not clickable due to scrollbar
### Prerequisites

### Description
When doing a find in project which generates a lot of results (enough to fill the window), you can't click on the last result because the (invisible) scrollbar is overlapping it.

### Steps to Reproduce
1. Search in project
2. Scroll to bottom to see last result
3. Click on result

**Expected behavior:**
Go to result

**Actual behavior:**
Nothing happens.

**Reproduces how often:**
100%

### Versions

OS X

Atom    : 1.20.0-dev-cb08c1b
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0

apm  1.18.3
npm  3.10.10
node 6.9.5 x64
python 2.7.10
git 2.14.1

### Additional Information
It looks like the problem is twofold: there is an (invisible) horizontal scrollbar, which is not needed in this case, and it overlaps the last result, making it unclickable.

Only seems to happen when you have this setting enabled in OS X:

image

When you set it to "Always", a smaller, dark atom themed horizontal scrollbar appears in the search results, but it sits below the results as expected.

When the option is set to "When scrolling", you can observe two different horizontal scrollbars, depending on where you are with your mouse when scrolling left/right. The normal one appears when you're scrolling the results, but an ugly, grey scrollbar appears when you scroll while hovering over the last result.

I've tried to capture this in a GIF:

https://www.dropbox.com/s/pk636pllc8p71o6/scrollbars.gif?dl=0

If you click just above the scrollbar, on the last result, there's about a 2 pixel capture area which does allow you to click on the last result.

simurai:
### Description of the Change

This PR makes sure there is no horizontal scrollbar so that the last result can be clicked.

The problem is that these `<div>`s have `width: 100%`.

image

And the parent `list-tree` has some padding. But the `width: 100%` of the children doesn't take the padding into account and so they extend to much, causing the horizontal scrollbar to appear.

**Fix 1**: Move the padding from the parent `list-tree` into the children. Now the `width: 100%` matches the parent's width.

find

**Fix 2**: Adds some bottom padding so there is always some space reserved for the horizontal scrollbar. I think it still looks ok.

find

### Alternate Designs

We could also make these `<div>`s not take up `100%` width, but stretch them with `left: 0 right: 0`.

### Benefits

Last result is clickable.

### Possible Drawbacks

None

### Applicable Issues

Closes #943

### DIFF
--- a/lib/project/list-view.js
+++ b/lib/project/list-view.js
@@ -84,6 +84,7 @@ module.exports = class ListView {
 
     return $.div(
       {
+        className: 'results-view-container',
         style: {
           position: 'relative',
           height: '100%',
--- a/styles/find-and-replace.less
+++ b/styles/find-and-replace.less
@@ -275,6 +275,13 @@ atom-workspace.find-visible {
     position: relative;
     flex: 1;
 
+    &-container {
+      // adds some padding
+      // so the last item can be clicked 
+      // when there is a horizontal scrollbar -> #943
+      padding-bottom: @component-padding;
+    }
+
     .list-tree .path {
       padding: 0 0 0 @component-padding;
     }
--- a/styles/find-and-replace.less
+++ b/styles/find-and-replace.less
@@ -241,6 +241,7 @@ atom-workspace.find-visible {
   position: relative;
   display: flex;
   flex-direction: column;
+  padding: 0;
 
   .preview-header {
     display: flex;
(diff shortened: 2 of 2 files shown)


adamreisnz:
Awesome thanks for the fix guys! 🎉

PR #994, opened 2018-01-19
Responsive results header

simurai:
### Description of the Change

- [x] This PR makes the results header responsive, so it can be used in more narrow panes or even as a dock item.
- [x] Reduces the indentation of the code in search results.

find

### Alternate Designs

Personally I'd like the "Collapse All" and "Expand All" just be a single toggle button. And also only have a single show/hide more lines number stepper that changes above and below.

### Benefits

Allows the search results be in a narrow place.

### Possible Drawbacks

None.

### Applicable Issues

This is a follow up to #992.

### DIFF
--- a/lib/project/results-pane.js
+++ b/lib/project/results-pane.js
@@ -93,7 +93,7 @@ class ResultsPaneView {
             {
               ref: 'previewControls',
               className: 'preview-controls',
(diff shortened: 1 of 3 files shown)


simurai:
Gonna merge even though it looks like Travis is stuck.

Issue #916, opened 2017-06-23
Find in project (multiple projects): Insertion of results does not preserve which files are collapsed


Mr-Wallet:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

When Atom has multiple projects imported, "Find in Project" searches the projects from top to bottom, not alphabetically. Consequently, results from alphabetically-prior projects lower in the list can insert into the middle of the Find In Project results, rather than appending to the end.

The issue is that which of those results are collapsed or not seems to be based on their position in the list (numerical index) rather than their path. This means that the wrong file can become collapsed if more search results come in after a file is collapsed.

This is particularly annoying when I'm trying to rule out irrelevant results by collapsing them. As it stands, I need to wait for my search to entirely finish before I can safely do so.

### Steps to Reproduce

1. Import two projects, making sure that the 2nd one in the Project sidebar is alphabetically before the 1st one. (I have only confirmed this with projects that are all within a common folder - in other cases, this bug may present differently if Atom is sorting results based on full, unqualified path names.)
2. Do a search that will get hits in both projects. After getting results from the first project but before getting all results from the 2nd project,
3. Collapse a file in the results (denote its position in the list as *X*).
4. After a result from the 2nd project is prepended to the list (because the project is alphabetically prior to the 1st project), position *X* is still collapsed, even though it now corresponds to a different file.

**Expected behavior:** Even as results come in, the collapsed entries should only be the files that were collapsed.

**Actual behavior:** Only the corresponding list positions remain collapsed, even when they now apply to different files than when that position was collapsed.

**Reproduces how often:** 100%

### Versions

```
atom --version

Atom    : 1.18.0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0
```
```
apm --version
apm  1.18.1
npm  3.10.10
node 6.9.5 ia32
python 2.7.10
[...]

PR #1000, opened 2018-03-12
Keep collapsed items in sync with search results

Lassi:
### Description of the Change

When searching across projects, if some items are collapsed in the search results view and new results are added, the collapsed items do not stay in sync with the results. Instead, collapsed items are "sliding" up and down as results are added/removed. This is due to the collapsed state being represented by an array of booleans that does not get updated when search results change. This commit introduces a fix to that issue by making sure that we update the aforementioned array as needed.

### Alternate Designs

An alternative to this design would have been to store the collapsed state directly on the element representing a "matched file" but it would entail a pretty big refactoring of the code which I don't feel I am familiar enough to start modifying in such length.

### Benefits

Whenever search results are updated, if some results were collapsed, these results will stay collapsed as expected. Previously, this "collapsed state" would be transferred to the search result above/below a collapsed item depending of the index where a search result would get inserted/removed.

### Possible Drawbacks

This might introduce some performance issues on very big sets of results as the array representing the collapsed items gets updated every time search results get updated.

### Applicable Issues

#916

### DIFF
--- a/lib/project/results-view.js
+++ b/lib/project/results-view.js
@@ -193,6 +193,11 @@ class ResultsView {
         this.selectedMatchIndex = selectedResult.matches.length - 1;
       }
     }
+
+    // Update the collapsedResultIndices to keep collapsed items in sync when a new result is added
+    if (filePathInsertedIndex != null) {
+      this.collapsedResultIndices.splice(filePathInsertedIndex, 0, false);
+    }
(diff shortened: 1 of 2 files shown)


PoignardAzur:
Damn, I was working on the same thing. (except my approach is the "refactor" one)

PR #1014, opened 2018-04-11
Check if an active editor exists when toggling "Only in Selection"

### LINKED ISSUE #1012: `Only in Selection` option with no open files throws an error
If you open a project and there are no open buffers, enabling `Only in Selection` in file find will generate the error below.

[Enter steps to reproduce:]

1. Open atom/atom in safe mode `atom --safe .`
2. Close any open buffers
3. `Find and Replace: Show`
4. Enable `Only in Selection`

Doesn't reproduce in 1.26.0-beta0, https://github.com/atom/find-and-replace/pull/1007 maybe related?

**Atom**: 1.27.0-dev-8ef1a05f2 x64
**Electron**: 1.7.11
**OS**: Mac OS X 10.12.6
**Thrown From**: Atom Core

### Stack Trace

Uncaught TypeError: Cannot read property 'getSelectedBufferRanges' of null

```
At /Applications/Atom.app/Contents/Resources/app/node_modules/find-and-replace/lib/find-view.js:712

TypeError: Cannot read property 'getSelectedBufferRanges' of null
    at FindView.toggleSelectionOption (/Applications/Atom.app/Contents/Resources/app/node_modules/find-and-replace/lib/find-view.js:712:37)
```

### Commands

```
     -0:05.8.0 core:close (input.hidden-input)
     -0:05.1.0 find-and-replace:show (atom-workspace.workspace.scrollbars-visible-always.theme-base16-tomorrow-dark-theme.theme-one-dark-ui)
```

### Non-Core Packages

```

```

thomasjo:
### Description of the Change

This resolves the null reference error reported in #1012 by simply checking if an active editor exists in the workspace. The bug was introduced in #1007, specifically [this change](https://github.com/atom/find-and-replace/pull/1007/files#diff-031d2cf6de79608fdbf0fde30f0fc2f8R701).

Note that I haven't found a simple way of implementing a regression test for this. I tried spying on `atom.onDidThrowError`, the notification system, etc. They work OK-ish when the test fails prior to applying this PR since I was using `waitsFor`. Although, that means once the fix has been applied, the `waitsFor` block in the regression test waits until timeout, which seems a bit unfortunate. Any proposals for writing a good regression test is very much appreciated.

### Alternate Designs

No other alternatives were considered.

### Benefits

Fixes an uncaught exception thrown by core functionality.

### Possible Drawbacks

None.

### Applicable Issues

Fixes #1012

### DIFF
--- a/lib/find-view.js
+++ b/lib/find-view.js
@@ -698,7 +698,9 @@ class FindView {
 
   toggleSelectionOption() {
     this.search({inCurrentSelection: !this.model.getFindOptions().inCurrentSelection});
-    if (this.model.getEditor().getSelectedBufferRanges().every(range => range.isEmpty())) {
+
+    let editor = this.model.getEditor();
+    if (editor && editor.getSelectedBufferRanges().every(range => range.isEmpty())) {
       this.selectFirstMarkerAfterCursor();
     }
   }


rsese:
Thanks @thomasjo :bow:

PR #1015, opened 2018-04-21
Add keymaps for home and end to project search

PoignardAzur:
### Description of the Change

Adds keymaps when pressing "home" and "end" in project search, that trigger `ResultsView.moveToTop` and `ResultsView.moveToBottom` respectively.

The code for these methods already existed, but it wasn't linked to any keymaps, and the only way to trigger them was to e.g. write "Core: Move To Top" in the command panel.

### Benefits

Being able to quickly select the first/last result in project search using the keyboard.

### Alternate Designs, Possible Drawbacks, Applicable Issues

No alternate design, no likely drawback, I don't think any issue was made on the subject.

### DIFF
--- a/keymaps/find-and-replace.cson
+++ b/keymaps/find-and-replace.cson
@@ -75,3 +75,9 @@
   'ctrl-enter': 'find-and-replace:replace-all'
 '.platform-linux .project-find .replace-container atom-text-editor':
   'ctrl-enter': 'project-find:replace-all'
+
+'.results-view':
+  'home': 'core:move-to-top'
+  'ctrl-home': 'core:move-to-top'
+  'end': 'core:move-to-bottom'
+  'ctrl-end': 'core:move-to-bottom'


PoignardAzur:
@Ben3eeE 

> Yeah I used the command from the command-palette. I wouldn't say the command is broken on master. It's just not bound to any keyboard shortcut by default. You could bind it in your personal keymap.

I don't know if it qualifies as broken then, but it's not ideal. Most users wouldn't even think to try typing "move-to-top/bottom", and would just assume the feature doesn't exist.

PoignardAzur:
@Ben3eeE 

> Yeah I used the command from the command-palette. I wouldn't say the command is broken on master. It's just not bound to any keyboard shortcut by default. You could bind it in your personal keymap.

I don't know if it qualifies as broken then, but it's not ideal. Most users wouldn't even think to try typing "move-to-top/bottom", and would just assume the feature doesn't exist.

(also, as a precedent, tree-view [does have default keybindings](https://github.com/atom/tree-view/blob/master/keymaps/tree-view.cson) for move-to-top/bottom)

lee-dohm:
Thanks very much for the help! :tada:

Issue #149, opened 2014-03-07
There's no way to exclude paths from a search


mattbasta:
When searching, filters can be specified. However, there's no way to exclude paths from the search. 

E.g.: `*.js, -*.min.js` cannot be used to exclude minified files from a search of JS

msisto:
+1

drw158:
It'd be great to get someone working on this. The solutions I see here don't apply to me.

If I'm doing a find and replace and want to exclude one of the files in the results, it's impossible. For example, if I want to replace `$light-blue` with `$blue` for every file *except* for `colors.scss`, it's not possible.

backspaces:
I'd like to eliminate "derived files" from my searches .. i.e. min.js, docco directory (documentation generated from src/ files), tools/ (.js files used in demos, not the project) and so on.  I'd like to do this on a per-repo basis.

Thus far it seems like atom does not search node_modules (and other files in my .gitignore file?).  I'm not sure why folks are seeing node_modules in their global searches .. they are typically not included in the repo, but referred to by package.json.

Apparently all I have to do is put the dirs I want to search in the "pattern" line of the global search pane.  It would be nice to automate that, somehow.  Not sure if best to have "positive" or "negative" i.e. include or ignore semantics.

   -- Owen

AGregg:
In the case of excluding a directory for a find and replace, my current workaround is to make sure my directory is clean, do the find and replace, then go into git bash and type `git checkout <directory>/.`

I would definitely like this feature to happen, but hopefully this can help people in the meantime.  @drw158

brendanfalkowski:
The ability to exclude folder's is very important to me. For Magento projects, there are sometimes 4 or 5 versions of the same file due to theme inheritance. Example:

```
/app/design/frontend/base/default/template/...
/app/design/frontend/default/default/template/...
/app/design/frontend/enterprise/default/template/...
/app/design/frontend/rwd/default/template/...
/app/design/frontend/custom/default/template/...

/skin/frontend/base/...
/skin/frontend/default/...
/skin/frontend/enterprise/...
/skin/frontend/rwd/...
/skin/frontend/custom/...
```

Only the `custom/default` files are being opened/edited 99% of the time.

**In Sublime Text,** I can exclude the other folders:

```javascript
{
"folder_exclude_patterns":
    [
        ".git",
        "*/frontend/base",
        "*/frontend/default",
        "*/frontend/enterprise",
        "*/frontend/rwd",
    ]
}
```

[...]

PR #1036, opened 2018-07-27
Add extra placeholder text on how to exclude files in the find-and-replace 3rd field

luzpaz:
### Description of the Change
When the `ctrl+shft+f` is pressed the search and replace feature opens. The screenshot below shows the field that I'm proposing to edit (3rd field that allows users to specify what path they prefer to search)
 
image

Currently the placeholder for the 3rd field reads as follows:
> File/directory pattern. eg. `src` to search in the "src" directory or `*.js` to search all JavaScript files

I'm proposing for the placeholder text to read as follows: 
> File/directory pattern. eg. `src` to search in the "src" directory; `*.js` to search all JavaScript files; `!*.png` to exclude all PNG files

_Note:_ The screenshot's function is to show where I mean to replace the placeholder text. I tried to actually manipulate the UI in the js console but atom reported an error so apologies for not demoing what it would look like. 

Ref: #149 

### Why Should This Be In find and replace?

I imagine that users wouldn't know that atom uses certain markup to dilienate (`!*.js`) what not to search. A user needs to dig around to find this out. I propose to educate the user on how to exclude files/directories in the placeholder text _directly_. I discovered this by searching the issue queue.  

### Benefits

Inform the user immediately through the placeholder how to achieve what they need

### Possible Drawbacks

Someone might view this as 'bloaty' or 'crowded' even though for me it fits perfectly all on one line

### DIFF
--- a/lib/project-find-view.js
+++ b/lib/project-find-view.js
@@ -103,7 +103,7 @@ class ProjectFindView {
             etch.dom(TextEditor, {
               ref: 'pathsEditor',
               mini: true,
-              placeholderText: 'File/directory pattern. eg. `src` to search in the "src" directory; `*.js` to search all JavaScript files; `!*.png` to exclude all PNG files',
+              placeholderText: 'File/directory pattern: For example `src` to search in the "src" directory; `*.js` to search all JavaScript files; `!*.png` to exclude all PNG files',
               buffer: this.pathsBuffer
             })
           )
--- a/lib/project-find-view.js
+++ b/lib/project-find-view.js
@@ -103,7 +103,7 @@ class ProjectFindView {
             etch.dom(TextEditor, {
               ref: 'pathsEditor',
               mini: true,
-              placeholderText: 'File/directory pattern: For example `src` to search in the "src" directory; `*.js` to search all JavaScript files; `!*.png` to exclude all PNG files',
+              placeholderText: 'File/directory pattern: For example `src` to search in the "src" directory; `*.js` to search all JavaScript files; `!src` to exclude the "src" directory; `!*.json` to exclude all JSON files',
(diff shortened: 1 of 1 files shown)


luzpaz:
Appveyor [fails due to another issue](https://ci.appveyor.com/project/Atom/find-and-replace/build/464/job/pluul5mvt60bpn3e#L7) unrelated to this PR

lee-dohm:
Would you mind changing the `eg.` here to `for example` while you're at it?

luzpaz:
@lee-dohm what about the period after `File/directory pattern.` ? Are you asking for me to remove that as well? 
`File/directory pattern for example src to search in the "src" directory`

luzpaz:
maybe using `!*.png` is not a great example? Perhaps using `!*.json` ? Also should I also mention a non-wild card negation example.. so using `!src` to exclude search the "src" directory ?

> File/directory pattern for example, `src` to search in the "src" directory; `*.js` to search all JavaScript files; `!src` to exclude searching in the "src" directory; `!*.json` to exclude all JSON files

luzpaz:
ping

luzpaz:
@lee-dohm thanks: 
What do you think about this revision?  
> File/directory pattern: For example `src` to search in the "src" directory; `*.js` to search all JavaScript files; `!src` to exclude the "src" directory; !*.json to exclude all JSON files.

lee-dohm:
@luzpaz Surround the `!*.json` with backticks (`) and it looks good :+1:

luzpaz:
@lee-dohm please review f1d89f2

lee-dohm:
Thanks for working through this with us @luzpaz :bow:

PR #1041, opened 2018-08-13
[WIP] add preserve case option including camel case, snake case, and kebab case

jedwards1211:
Happy birthday folks!  🎉 
I need to add tests still.

### Requirements

### Description of the Change

This adds an option to preserve case when replacing.  It's very comprehensive, for instance replacing `foo bar` with `hello world` will make the following replacements, more than any other editor I've used is capable of:

```
foo bar -> hello world
Foo Bar -> Hello World
Foo bar -> Hello world
FOO BAR -> HELLO WORLD
fooBar -> helloWorld
FooBar -> HelloWorld
foo-bar -> hello-world
Foo-Bar -> Hello-World
foo_bar -> hello_world
FOO_BAR -> HELLO_WORLD
```

It can preserve case equally well when replacing regular expressions, though the matches will be limited by the regexp.  For instance, replacing `(foo) (bar)` with `$2 $1` will make the following replacements:

```
foo bar -> bar foo
Foo Bar -> Bar Foo
Foo bar -> Bar foo
FOO BAR -> BAR FOO
```

And replacing `(baz)[-_ ]?(qux)` with `$2 $1` will make these replacements:
```
baz qux -> qux baz
Baz Qux -> Qux Baz
Baz qux -> Qux baz
BAZ QUX -> QUX BAZ
bazQux -> quxBaz
BazQux -> QuxBaz
baz-qux -> qux-baz
Baz-Qux -> Qux-Baz
baz_qux -> qux_baz
BAZ_QUX -> QUX_BAZ
```

### Alternate Designs

The implementation uses my `preserve-case` package.
I have seen other packages for doing this, but none are as comprehensive.

### Benefits

Case-preserving replace can be extremely useful for programmers.  Often the same term will appear in multiple cases in code.  For instance in `find-view.js` you can find `wholeWord`, `WholeWord`, and `whole-word`.  Being able to replace all of these in one fell swoop without botching the case is very convenient.

### Possible Drawbacks

At the moment, `foo_bar` won't find `foo_Bar` when preserve case is turned on, because it only searches for the supported case types rather than doing a true case-insensitive search.

Turning preserve case on may actually increase find results since `hello world` will additionally match `hello-world`, `hello_world`, and `helloWorld`.

Using this with regexes can be a bit counterintuitive, because it performs the regex substitution first, then restores the original case of the match, which may end up adding or removing a space or `-` or `_`, regardless of what was in the replacement pattern.

### DIFF
--- a/keymaps/find-and-replace.cson
+++ b/keymaps/find-and-replace.cson
@@ -38,6 +38,7 @@
   'cmd-alt-c': 'find-and-replace:toggle-case-option'
   'cmd-alt-s': 'find-and-replace:toggle-selection-option'
   'cmd-alt-w': 'find-and-replace:toggle-whole-word-option'
+  'ctrl-alt-cmd-c': 'find-and-replace:toggle-preserve-case-option'
 
 '.platform-win32 .find-and-replace, .platform-linux .find-and-replace':
   'shift-enter': 'find-and-replace:show-previous'
@@ -45,6 +46,7 @@
   'alt-enter': 'find-and-replace:find-all'
   'ctrl-alt-/': 'find-and-replace:toggle-regex-option'
   'ctrl-shift-c': 'find-and-replace:toggle-case-option'
+  'ctrl-alt-shift-c': 'find-and-replace:toggle-case-option'
 
 '.platform-darwin .project-find':
   'cmd-enter': 'project-find:confirm'
--- a/lib/buffer-search.js
+++ b/lib/buffer-search.js
@@ -1,6 +1,7 @@
 const { Point, Range, Emitter, CompositeDisposable, TextBuffer } = require('atom');
 const FindOptions = require('./find-options');
 const escapeHelper = require('./escape-helper');
+const replace = require('preserve-case')
 
 const ResultsMarkerLayersByEditor = new WeakMap;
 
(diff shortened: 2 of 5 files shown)


lee-dohm:
Thank you very much for your contribution.

Unfortunately, as I stated in https://github.com/atom/find-and-replace/issues/165, this functionality isn't something that we want to maintain long-term. We can understand that this would be very useful for certain workflows, however it isn't something that we're going to be able to take on.

Thank you very much for the work you put into this and your interest in Atom.

jedwards1211:
Okay.  What do you think would be best in the interest of making this something convenient for others to use - that I release it as a replacement package for find-and-replace under a different name, or do you think find-and-replace is pluggable enough I could somehow release a plugin-plugin for it?

DamnedScholar:
Something that would be easier to maintain is to [emit an event](https://atom.io/docs/api/v1.29.0/Emitter) when text is replaced. The payload could be in the form `{head: [4, 10], old: "fooBar", new: "barfoo"}`, which would provide all the necessary information for *any* package to listen and then act on a replacement action.

jedwards1211:
Webpack uses [tapable](https://github.com/webpack/tapable) which provides a more structured way of creating hooks for plugins.

DamnedScholar:
It would manipulate the buffer. The purpose of the event would just be to give out a location and what's been replaced with what.

jedwards1211:
@DamnedScholar ah, I see.  In the case of regexps would `new` be the regexp the user entered, the result of `regexp.exec`, only the match text, or?

DamnedScholar:
If you're just emitting a replacement event, the basic information you need to know is what the old text was and what the new text is. If you want to preserve the regular expression, that would be an additional key in the event object.

Issue #653, opened 2016-01-25
Find and Replace in large buffer is very slow


benogle:
_From @ypresto on January 21, 2016 7:59_

It seems to rebuilding tokenized buffer on every operation in transaction.

![image](https://cloud.githubusercontent.com/assets/400558/12474497/019feaf4-c05f-11e5-94b9-1e07fd3b66fa.png)

It took 16 secs to replace 920 `\n`s to newline on 15-inch MacBook Pro Retina Late-2013.

_Copied from original issue: atom/atom#10531_

benogle:
This is actively being worked on by allowing multiple updates to update the interface and any listeners once. iirc, we do not have an issue to link to at this time. @nathansobo are there any issues related to this?

lllShamanlll:
Are there any movements here? Didn't find any related atom issue.

nathansobo:
There was improvement following the issue, but it's always a tricky issue to discuss when performance is "good enough" to close an issue. My answer is that it probably still isn't good enough and we're focusing on some other areas for now. We will revisit find and replace performance in the future.

nathansobo:
That said, if you'd like to post profile data on any performance issue it is always welcome.

lllShamanlll:
Actually this looks like atom issue.  Due to lack of methods to work with multiple selections, it requires to add selections one by one(what find-and-replace does) & as far as merge performs on every addition, that cause O(N^3) complexity to select-all, where N - count of selections.

So now it takes about 30sec. to select all new-lines in file, that contain 1000 empty lines.
Is that "good enough"? In SBT3 it takes less than a second for instance.
![screenshot from 2016-03-25 17 03 05](https://cloud.githubusercontent.com/assets/840714/14045521/035a594a-f2ac-11e5-8d24-ef99ccebc08f.png)

nathansobo:
It's definitely not good enough, though opening a specific issue addressing `select-all-occurrences` and posting your saved timeline data would be helpful. We have some challenges due to the synchronous nature of Atom's editor APIs. We need to process selections in a batched manner.

matthew-dean:
This is still a severe issue on version 1.12.7 of Atom and 0.202.2 of find-and-replace. I would replace "not good enough" with "probably the worst large file loading and searching of any JS-based editor". https://github.com/ajaxorg/ace handles large files, loading and searching, very smoothly.

jpike88:
[...]

PR #1048, opened 2018-09-10
Improve efficiency of replace algorithm

jedwards1211:
### Description of the Change

The current code for `replace` is O(n^2), where n is the number of markers: for each marker, it does an O(n) search for that marker in `this.markers`, and then does an `O(n)` splice to remove that marker.

I have re-sanctified it by creating a `Set` of the markers that will be replaced (O(n)), replacing the markers (O(n), ignoring the fact that `this.editor.setTextInBufferRange` is probably not truly constant time), and then filtering out the markers that are in the set (O(n)).

### Alternate Designs

A more proper solution would use a `Set` or `Map` (or maybe multiple indexes) for `this.markers`.  This is an interim solution that creates a temporary `Set` to make removing the replaced markers more efficient.

### Benefits

The asymptotic efficiency of `replace` goes from O(n^2) to O(n).

### Possible Drawbacks

Laymen who have not be initiated into the mystic wisdom of big-O notation may be baffled by this change.  As with any good religion, we will have to exhort them to accept it upon faith.

### Applicable Issues

#653

### DIFF
--- a/lib/buffer-search.js
+++ b/lib/buffer-search.js
@@ -107,7 +107,8 @@ class BufferSearch {
       }
 
       function removeInPlace(array, setToRemove) {
-        for (let src = 0, dest = 0, end = array.length; src < end; src++) {
+        let dest = 0
+        for (let src = 0, end = array.length; src < end; src++) {
           const elem = array[src];
           if (!setToRemove.has(elem)) array[dest++] = elem;
         }
--- a/lib/buffer-search.js
+++ b/lib/buffer-search.js
@@ -95,8 +95,6 @@ class BufferSearch {
         replacePattern = escapeHelper.unescapeEscapeSequence(replacePattern);
       }
 
-      const replacedMarkers = new Set(markers);
-
       for (let i = 0, n = markers.length; i < n; i++) {
         const marker = markers[i];
         const bufferRange = marker.getBufferRange();
@@ -107,7 +105,17 @@ class BufferSearch {
 
         marker.destroy();
       }
-      this.markers = this.markers.filter(marker => !replacedMarkers.has(marker));
+
+      function removeInPlace(array, setToRemove) {
+        for (let src = 0, dest = 0, end = array.length; src < end; src++) {
+          const elem = array[src];
+          if (!setToRemove.has(elem)) array[dest++] = elem;
+        }
+        array.length = dest;
+      }
+
+      const replacedMarkers = new Set(markers);
+      removeInPlace(this.markers, replacedMarkers);
     });
 
     return this.emitter.emit('did-update', this.markers.slice());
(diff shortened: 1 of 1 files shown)


50Wliu:
> A more proper solution would use a Set or Map (or maybe multiple indexes) for this.markers. This is an interim solution that creates a temporary Set to make removing the replaced markers more efficient.

Seems like this should be a code comment as well, so it's not forgotten after this PR is merged.

lee-dohm:
@jedwards1211 a part of your PR body was deleted as a violation of the [Atom Code of Conduct](https://github.com/atom/atom/blob/master/CODE_OF_CONDUCT.md) as it was insulting or derogatory. You may consider this an official warning.

jedwards1211:
@50Wliu well, I think it bears more investigation in general.  For instance, I'm not sure if it's actually worth it for the `MarkerIndex` to be a tree, since building that tree takes `O(n log n)` time and lookups take `O(log n)` time, where `n` is the number of markers.  I wonder if using a hash that can tell us if there's a marker at a given start index would perform better in general, since it could be built in `O(n)` time, and lookups are `O(1)` (times the number of characters visible on screen, which is much more bounded that the number of markers found).  The only thing that a tree is better at in all cases is determining which markers are within the selection, which is `O(log n)`, versus a hash which would be `O(n)`.

maxbrunsfeld:
@jedwards1211 I agree this code is pretty naive; it's been around for a really long time. It looks like what you did *might* be a useful optimization, but it also might not. With any optimization, before you engage in much speculative discussion about algorithm complexity, you should do this:

1. Record a profile of the operation in question.
2. Make your change.
3. Record a new profile of the operation.
4. Compare the two profiles.

Before speculating further, can you please perform these steps?

maxbrunsfeld:
Also, unfortunately this PR has nothing to do with the slowness documented in #653. That is due to the inefficient way that we pass changes from the TextBuffer to the language mode in order to update syntax highlighting. It is on my personal roadmap to fix. The first step of it is here: https://github.com/atom/text-buffer/pull/279. Unfortunately, I haven't had time to work on this lately.

jedwards1211:
@maxbrunsfeld You are right, I should have profiled it.  I ended up writing a separate simulation to profile removing a million integers from a 10-million element array using the `.indexOf`/`.splice` method and the `Set` method, since (as you mentioned) `setTextInBufferRange` is taking the majority of the time there.  It only took about a second for the naive code to do that, and only a few 100 ms less for the `Set` method.  And while I do expect the cost of the naive method to increase as O(n^2) (that can be derived mathematically) the size of the matches array would exhaust available memory before it gets to the point it could take much more than 10 seconds to do the array removals on a modern machine.
So, I'm sorry for the shaming.

I had another idea for an interim change that might actually cause a substantial improvement now, while we wait for your `text-buffer` PR: building the result text for the entire buffer, and then setting the buffer text in one go.  I'll experiment with that and if it seems useful I'll make that into a PR.

PR #1061, opened 2018-12-19
Remove hardcoded fonts so that search view always use Atom’s font

### LINKED ISSUE #1057: Results don’t use the same font as in other source code views
*Edit by @rsese - to add notes from https://github.com/atom/find-and-replace/issues/1057#issuecomment-438810270*

Looks like the behavior changed in 1.28.0, here's how things looked in 1.27.2, the code uses the font I set while the filename headers don't:

But in 1.28.0 it swapped, the font I set is used for the filename header and the code doesn't:

And this is how things look on 1.32.2 with macOS 10.12.6.

Maybe related to https://github.com/atom/find-and-replace/pull/1002.

---

 ### Prerequisites

### Description

Find and replace results aren’t displayed with the same font as configured in Editor → font-family

### Steps to Reproduce

1. Set a distinguishable font
2. Launch a search

**Expected behavior:** find and replace text results uses the same font as for other source code views.

**Actual behavior:** find and replace text results uses a different font than in other source code views.

See how the file name (using Atom’s font) is different from the source code extract which is using another font I didn’t configure:
screenshot_20181113_184707

**Reproduces how often:** all the time

### Versions

On Arch Linux:

```
Atom    : 1.32.2
Electron: 2.0.12
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

ariasuni:
### Description of the Change

Remove the hardcoded font names, so that the package use Atom’s font.

Before (setting the Editor font to Inconsolata):
screenshot_20181220_052703

After:
screenshot_20181220_052831

### Alternate Designs

Expose the font used for file names as a configuration option, but I fail to find a compelling use case.

### Benefits

Less code, uniform look, don’t need to change several fonts.

### Possible Drawbacks

Can’t use another font for file names in search results.

### Applicable Issues

Fix #1057

### DIFF
--- a/styles/find-and-replace.less
+++ b/styles/find-and-replace.less
@@ -1,8 +1,6 @@
 @import "ui-variables";
 @import "syntax-variables";
 
-@find-and-replace-font: Menlo, Consolas, 'DejaVu Sans Mono', monospace;
-
 // result markers
 atom-text-editor {
   .find-result .region {
@@ -311,7 +309,6 @@ atom-workspace.find-visible {
       margin-right: 1ex;
       text-align: right;
       display: inline-block;
-      font-family: @find-and-replace-font;
     }
     .match-row.selected .line-number {
       color: @text-color-selected;
@@ -324,7 +321,6 @@ atom-workspace.find-visible {
 
     .preview {
       word-break: break-all;
-      font-family: @find-and-replace-font;
       white-space: pre;
       color: @text-color-subtle;
     }


rsese:
Thanks @ariasuni!  Can you add before and after screenshots that show the fix?

ariasuni:
I updated my first comment with screenshots. :slightly_smiling_face:

rsese:
Thank you! Someone from the team will take a look as soon as they can.

daviwil:
Going to merge this anyway as I think it's a useful change.  Thanks again @ariasuni!

simurai:
Sorry for the late reply, but yes.. this is a great fix. 👍 👍 👍  Thanks @ariasuni

PR #1064, opened 2018-12-28
Add ability to pull selected text into replace field

50Wliu:
### Requirements

### Description of the Change

This is an update of #719.

### Alternate Designs

None.

### Benefits

Allows the replace field to be easily populated with the selected text.

### Possible Drawbacks

People may already be using <kbd>ctrl-shift-e</kbd> for other purposes.

### Applicable Issues

#67.
Supersedes and closes #719.


lib/find-view.js
```diff
@@ -593,9 +594,27 @@ class FindView {
       if (this.model.getFindOptions().useRegex) {
         findPattern = Util.escapeRegex(findPattern);
       }
+
       if (findPattern) {
         this.findEditor.setText(findPattern);
-        this.search();
```
daviwil: Looks like this no longer activates the search immediately, is that the desired behavior?

50Wliu: Ah you're right, I thought this wasn't doing anything useful, but it turns out it does have a purpose if live search is disabled. I'll push a new commit fixing that.

daviwil:
Thanks @50Wliu!  Looks good, merging it.

Issue #1084, opened 2019-05-22
Searching on a modified buffer can lead to broken UI


rafeca:
### Description

When searching using multiline regexps with a modified file opened, the search results get completely broken:

<img width="880" alt="Screenshot 2019-05-22 at 17 22 12" src="https://user-images.githubusercontent.com/408035/58193401-315a9e00-7cc3-11e9-93b4-a4a6890cef52.png">

### Steps to Reproduce

1. Open a repo with e.g a `package.json` file on Atom
2. Open the `package.json` file and do some modifications on it without saving them.
3. Open find and replace and search for the following regexp: `,\n\s+`

**Expected behavior:** Get some results, e.g:

<img width="754" alt="Screenshot 2019-05-22 at 18 26 09" src="https://user-images.githubusercontent.com/408035/58193491-66ff8700-7cc3-11e9-80e8-0f22a962633e.png">

**Actual behavior:** The UI gets broken

**Reproduces how often:** 100%

### Versions

All versions

Issue #1075, opened 2019-03-19
Improve time to find on a large repository by using `ripgrep`


rafeca:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

After seeing the big benefits that `ripgrep` has brought to `fuzzy-finder` ([more info](https://github.com/atom/fuzzy-finder/pull/369)), we want to explore what would be the impact on the time it takes to find something on a large project (which is not specially fast at the moment).

BurntSushi:
Please let me know if there's anything I can help with here. Note that ripgrep does have a JSON output mode which should make this task much easier to get right, but its documentation might not be in an obvious place. Here's a link: https://docs.rs/grep-printer/0.1.1/grep_printer/struct.JSON.html

rafeca:
Thanks @BurntSushi ! Really excited to integrate `ripgrep` in Atom and get some perf boosts!

I'll take a look at the JSON output docs

rafeca:
We have a PR in `atom/atom` which adds `ripgrep` as a backend system for the `scanner`: https://github.com/atom/atom/pull/19348

Once that one gets merged we're going to be able to use it from this package 🎉

rafeca:
Closing this issue, we're going to have an option for `ripgrep` powered search on Atom v1.39, and hopefully we're going to make it the default one very soon.

Thanks @BurntSushi for such an awesome tool! it's been a delight to integrate it on Atom ❤️

Issue #303, opened 2014-10-24
Can't search across lines with .* regex


izuzak:
_Originally reported by @Cydrobolt over at https://github.com/atom/atom/issues/3892_

---

Both regex and normal find can't search across lines. E.g

```

something="ex
ists"


<p>blablablabla
blablablabla</p>
```

For the second example, a regex of `<p>.*</p>` should have matched the text. However, it does not work, because it is spread across two lines.

rnyberg:
Ubuntu 14.04 crashs still exists.

harai:
It still crashes.

* Ubuntu 15.04
* Atom 1.0.0
* find-and-replace 0.174.1

impinball:
To be honest, I don't think that crash is easily fixable. Try running that regex through `grep` and see what happens. If it doesn't hang on a file of about 30 lines, then there is likely a very difficult perf bug in V8 or (highly unlikely) Atom's text editor. I would be surprised if that's the case, though, considering that is literally "any number of a group of the least number of characters consisting of either a newline or the largest group of non-line-ending characters you can get". That's a lot of work to do, and it's not the easiest to even statically compile that regex to infer that it's matching any set of characters that don't include any line-ending character other than a line feed. That means the regex doesn't match other line-breaking characters, i.e. carriage return, (the obscure line-breaking code points) U+2028 and U+2029, etc. Another thing is that even Sublime, etc. tend to choke a little on regular expressions.

Regular expression engines are extremely slow to begin with, and V8's Irregexp engine is one of the few that isn't atrociously slow. (It's faster than most POSIX-based regex implementations, and it's faster than Perl's highly optimized, highly flexible one.)

I would say one way, probably the best way, to curb the crashes is to instate a delay since the last character is added before the regex is finally executed, even as little as 200 milliseconds. I couldn't tell you how many times I've had Atom crash in the middle of me typing out a regex, simply because the incomplete one happened to match a third of the code. The other thing is that most editors don't regular expressions as they're typed - they run via a dialog or similar. Atom is rather unique in this problem.

acusti:
I don’t know if this is worth it’s own issue, but the general problem of not being able to do a multiline search without converting your search to a regular expression is a painful one. It seems like the need to use “replace in project” to modify every instance of a multiline chunk of code in a project is the kind of thing that comes up frequently enough that it would be great if the editor could handle it. There have been a few times that I just wanted to paste in a code chunk to the find field and a different one in the replace field.

[...]

PR #1085, opened 2019-05-22
Handle multiline results on the find and replace UI

rafeca:
### Description of the Change

This PR tweaks the logic of the method that handles results from the search provider to be able to support results that span across multiple lines.

### Motivation

The current default scanner ([`scandal`](https://github.com/atom/scandal)) does not support multiline regexps, so up until now it was rare to get a search result that contained multiple lines, this is why the current UI does not handle them well.

Still, currently there's one situation where multi-line results are shown, which causes a very broken results view ([more info](https://github.com/atom/find-and-replace/issues/1084)), so this PR aims to fix it.

This PR also paves out the support for [adding `ripgrep` as backend for the searches](https://github.com/atom/find-and-replace/issues/1075), since `ripgrep` can return multi-line results (this is going to be added in a separate PR).

### Benefits

This will allow find and replace to show multiline results, which is a much requested feature: https://github.com/atom/find-and-replace/issues/303

### Possible Drawbacks

Slightly more complex logic for displaying results.

### DIFF
--- a/lib/project-find-view.js
+++ b/lib/project-find-view.js
@@ -318,7 +318,7 @@ class ProjectFindView {
     return searchPromise;
   }
 
-  search(options) {
+  async search(options) {
     // We always want to set the options passed in, even if we dont end up doing the search
     if (options == null) { options = {}; }
     this.model.getFindOptions().set(options);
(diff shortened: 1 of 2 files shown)


Abby199:
Duplicado de # 400

Issue #806, opened 2016-11-02
"Ignored names" from global settings are not handled correctly


mik01aj:
My company uses a proprietary build system `adslkfsdjk`. It produces lots of files in directories named `adslkfsdjk-genfiles`, `adslkfsdjk-out`, `adslkfsdjk-testlogs`, etc. I added `adslkfsdjk-*` as ignored in Atom settings, and they disappeared from tree view. Good!

But when I search for something using `find-and-replace`, it seems to search all the files (including these), because it hangs forever showing an empty result list with this: 

![image](https://cloud.githubusercontent.com/assets/3027415/19929500/f21e6f84-a103-11e6-9a94-14ae277cf8aa.png)

...and my project has only around 2000 files. :open_mouth: 

Note that **when I put a list of these folder names explicitly in "Ignored names", the search starts working** (and it's reasonably fast). Seems like a bug to me. :bug:

mik01aj:
One more: When I specify a file/directory pattern like `**/detail/*` in find-and-replace, it starts searching these ignored folders again. :confused:

joseramonc:
Agree, it's happening to me too, It would be nice to have an option to ignore files in global search like fuzzy-finder does.

![screen shot 2016-12-09 at 11 50 41 am](https://cloud.githubusercontent.com/assets/1993929/21058898/bb19c1fe-be05-11e6-8de6-f7beb3c6b93e.png)

chrishandorf:
Me too.  I don't want to search the node_modules subfolder when searching within my project.

dpehrson:
Also hitting this. I'm confused why find-and-replace searches VCS ignored folders even though that config specifically says it won't?

dwelle:
The `Ignored names` setting doesn't say it will be respected by `find-and-replace` package.

WTBS, it'd be nice to have a per-project settings of ignored paths for `find-and-replace`, possibly specified in a special file at the root of the project (I specifically _don't want_ to put the paths into `.gitignore`).

Issue #571, opened 2015-10-22
No support for lookbehind


naomik:
Implementing JavaScript's native RegExp engine for a find/replace tool in a text editor seems like a dramatic oversight.

As find-and-replace leans on JS's RegExp, of course the following find expressions is invalid

``` js
(?<=hello)world
(?<!hello)world
```

Errors

``` plain
Invalid regular expression: /(?<=hello)world/: Invalid group
Invalid regular expression: /(?<!hello)world/: Invalid group
```

This module is unusable without such features

Epigene:
I agree, lookaround is crucial!

Using JS's RegExp is also weird considering Oniguruma is already used within Atom to parse grammar files as mentioned in [this exchange](https://discuss.atom.io/t/regular-expression-lookaround-support-in-find/12400).

abe33:
I as already mentioned on Discuss, using oniguruma for grammar made sense as it allows to reuse TextMate and SublimeText grammars out of the box, but AFAIK that's the only context where oniguruma is used. 
OTOH I agree it's a shame that lookbehind is not part of JS regexp (while loohahead is).

JesseLeite:
I don't know much about the subject, apart from lookbehinds not being part of JS regex.  However, as a PHP dev who regular-ily (pun intended) regexes via PHP regex  engine with lookbehinds, it sucks not being able to lookbehind in Atom's find and replace.  Curious about this Oniguruma?

alexchandel:
Oniguruma has this feature and many, many more. There is absolutely no reason to use JS's garbage regex engine over Oniguruma.

JesseLeite:
@alexchandel I'm unfamiliar with Oniguruma. I'm all for more features, but is the syntax standard and testable on a site like [regex101.com](http://regex101.com)?

50Wliu:
@JesseLeite It's what Atom uses for languages, and I've never had a problem testing any of those regexes on regex101.com.

bb-jakespirek:
Bummed that lookbehind doesn't work.

alexchandel:
@bb-jakespirek It would if #698 happened

alexchandel:
Since #698 is closed as a duplicate, is this issue now for using Oniguruma for regexes?

naomik:
relevant: https://github.com/atom/find-and-replace/issues/570

Oniguruma would fix this too.

bric3:
@JesseLeite https://regex101.com/r/dO0yG2/1

[...]

PR #1086, opened 2019-05-24
Add config option to use `ripgrep` for scanning files

rafeca:
## Summary

This PR adds a config option to enable `ripgrep` powered find and replace on this package. The integration with `ripgrep` has been done in Atom core on https://github.com/atom/atom/pull/19348 and this PR just opts in the find and replace package to use it.

Screenshot 2019-05-24 at 12 14 38

## Benefits

Using `ripgrep` speeds up drastically the time to find files on any kind of repository (we're taking about **up to 22X faster times**):

| Type | Num Files | Time (standard)  | Time (ripgrep) | Improvements  |
|---|---|---|---|---|
|  Small | 2K  | 940ms  | 62ms | **15X** faster  🎉 |
|  Medium |  30K | 7.7s  | 620ms  | **12X** faster 🎉 |
|  Large (returning 5 results) | 270K  |  129s  |  5.9s  |  **22X** faster 🎉 |
|  Large (returning 26k results) | 270K  |  142s  |  17.5s  |  **8X** faster 🎉 |

*(the last measure has been done to check the less favourable case for `ripgrep`, which is a search that returns a lot of results (26k) which need to be passed from the `ripgrep` process to Atom. This is still 8X times faster than the current search logic. We could limit the number of results, but this can be done as a separate PR).*

## Possible Drawbacks

The search behaviour with `ripgrep` is slightly different than the one currently implemented, and there may be still some edge cases to polish. There's more information about the changes in https://github.com/atom/atom/pull/19348.

To mitigate that, I've done quite extensive tests but I'll keep checking for potential edge cases. Also, I'm planning to leave this config flag at least for 1 version of Atom so we can catch as many issues as possible before shipping this to users by default .

## Applicable Issues

https://github.com/atom/find-and-replace/issues/1075

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -17,6 +17,7 @@ Currently Find and Replace does not log any counter events.
   |-------|-------|
   | `ec` | `time-to-search`
   | `ev` | Number of found results
+  | `el` | Search system in use (`ripgrep` or `standard`)
 
 ## Standard events
 
--- a/lib/project/results-model.js
+++ b/lib/project/results-model.js
@@ -216,6 +216,7 @@ module.exports = class ResultsModel {
     const trailingContextLineCount = atom.config.get('find-and-replace.searchContextLineCountAfter')
 
     const startTime = Date.now()
+    const useRipgrep = atom.config.get('find-and-replace.useRipgrep')
 
     this.inProgressSearchPromise = atom.workspace.scan(
       this.regex,
@@ -223,6 +224,7 @@ module.exports = class ResultsModel {
         paths: searchPaths,
         onPathsSearched,
         leadingContextLineCount,
+        ripgrep: useRipgrep,
         trailingContextLineCount
       },
       (result, error) => {
@@ -244,7 +246,11 @@ module.exports = class ResultsModel {
     } else {
       const resultsSummary = this.getResultsSummary()
 
-      this.metricsReporter.sendSearchEvent(Date.now() - startTime, resultsSummary.matchCount)
+      this.metricsReporter.sendSearchEvent(
+        Date.now() - startTime,
+        resultsSummary.matchCount,
+        useRipgrep ? 'ripgrep' : 'standard'
+      )
       this.inProgressSearchPromise = null
       this.emitter.emit('did-finish-searching', resultsSummary)
     }
--- a/lib/reporter-proxy.js
+++ b/lib/reporter-proxy.js
@@ -19,10 +19,11 @@ module.exports = class ReporterProxy {
     delete this.reporter
   }
 
-   sendSearchEvent (duration, numResults) {
+   sendSearchEvent (duration, numResults, crawler) {
     const metadata = {
       ec: 'time-to-search',
-      ev: numResults
+      ev: numResults,
+      el: crawler
     }
 
      this._addTiming(duration, metadata)
--- a/package.json
+++ b/package.json
@@ -125,6 +125,12 @@
       "title": "Show Search Wrap Icon",
       "description": "Display a visual cue over the editor when looping through search results."
     },
+    "useRipgrep": {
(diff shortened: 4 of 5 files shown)


rafeca:
Going to merge this one, if there's any feedback afterwards I'll handle it on a separate PR

rafeca:
As @maxbrunsfeld has mentioned on Slack, the `ripgrep` scanner also fixes https://github.com/atom/find-and-replace/issues/806, which is a long-standing issue regarding not ignoring VCS-ignored files when there's a filter path in the project search.

dwelle:
Scanning the commit, it'd be great to allow to enable ripgrep's PCRE2 so that lookbehind assertions work, too (https://github.com/atom/find-and-replace/issues/571). AFAIK it has a small perf regression, so if that's a concern, best adding it as opt-in only (ala vscode).

rafeca:
@dwelle thanks for the suggestion!

Would you be willing to contribute adding this as a config option?

dwelle:
Sure, provided it's not time-sensitive because I reckon it'll take a while --- I'll have to set up a dev enviro and figure out how to run it (plus I have little time ATM, but who does). Alternatively, I can write the code + tests blindly, cross my fingers and let the CI and reviewers do the rest :).

Btw, should a new [event metadata field](https://github.com/atom/find-and-replace/blob/8f0aa2d4785f378f3516573208dbbc18f15d7d70/docs/events.md#timing-events) be added, too?

rafeca:
Thanks @dwelle! I don't think time sensitive, so it's fine 😄 

Setting up the dev environment is quite easy, you can follow the instructions [here](https://flight-manual.atom.io/hacking-atom/sections/contributing-to-official-atom-packages/) and [here](https://flight-manual.atom.io/hacking-atom/sections/hacking-on-atom-core/), but if you prefer you can let the CI do the job 😭

Just take into account that the search logic is [in Atom core](https://github.com/atom/atom/blob/master/src/ripgrep-directory-searcher.js#L271). You can add a new option to the [`workspace.scan()` method] (https://github.com/atom/atom/blob/master/src/workspace.js#L2048) to use PCRE2, and then pass it from `find-and-replace` based on the config option that you create (as [I did for `ripgrep`](https://github.com/atom/find-and-replace/pull/1086/files#diff-d8fbdddaf0ac028a971cce25034a773a)).

You don't need to care about the `DefaultDirectorySearcher` implementation (just mention on the configuration option description that it only works on `ripgrep`).

Also, make sure to create a new test on the `workspace-spec.js` file that asserts that the PCRE2 option works well (you can use [any existing test as inspiration](https://github.com/atom/atom/blob/master/spec/workspace-spec.js#L2745-L2755)).

Regarding the metadata field, I don't think it's needed for now. Only if we see that the performance of the searches regress after this change we'll add that.

Thanks again for your help!

dwelle:
@rafeca finally managed to free up some time to work on this.

Being on Win7 I've been trying to bootstrap the atom repo for quite some time but it keeps failing on the well known  `msbuild.exe failed with exit code: 1` error.

I believe I have the correct version of `Visual C++ Build Tools`, but as for the `Windows 8 SDK`, not sure. The installation shows as successful, but then doesn't show as installed when trying to install it again (to verify).

I'll try to re-install everything, but if I don't break through soon, I'll have to go with my original idea and write the code blind.

witygass:
Yaaaaas! Thank you so much. Alleviates one of the prime badgering points that my co-workers love to make fun of me for when I say that I still use atom after years of being a dev.

popovba08:
มันยังไงกันนะ ระบบออโต้ ที่เร็วที่สุดในโลก

มั่นคง ปลอดภัย ไว้ใจ [berlin777](https://berlin777.com/)

PR #1091, opened 2019-06-20
Construct ResultsView directly to fix some flaky tests

nathansobo:
Fixes https://github.com/atom/atom/issues/19480

Previously, all tests on the `ResultsView` actually ran a real search against the file system to populate their results. We're seeing intermittent failures in a few tests, so I decided to eliminate the async behavior by constructing a `ResultsView` directly with a model and then populating some fake results on the model. This makes the test synchronous and should hopefully eliminate the flakiness.

### DIFF
--- a/lib/project/results-model.js
+++ b/lib/project/results-model.js
@@ -387,3 +387,6 @@ module.exports = class ResultsModel {
     return pathsPattern.trim().split(',').map((inputPath) => inputPath.trim())
   }
 }
+
+// Exported for tests
+module.exports.Result = Result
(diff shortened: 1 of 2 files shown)


jasonrudolph:
Nice work, @nathansobo! Do you think a similar approach would help with the `ResultsView` flake seen in https://github.com/atom/atom/issues/19554?

nathansobo:
Sorry I missed this question. I ended up making another pass in #1093 that keeps most of the tests async but removes some of the code that I think was causing flakiness.

PR #1095, opened 2019-06-30
Add pcre2 ripgrep option

dwelle:
### Description of the Change

Added option to enable `pcre2` ripgrep regex engine, as discussed [in here](https://github.com/atom/find-and-replace/pull/1086#issuecomment-502605862).

**NOTE:** I haven't written any specs because I'm not sure what steps to take. Should I wrap all the specs in additional loop, same as [prev PR has done with ripgrep support](https://github.com/atom/find-and-replace/pull/1086/files#diff-3529658c67b9493659c04f7644950802R12)?

**NOTE:** cannot build Atom ATM so haven't tested the changes locally.

### Benefits

Allows additional regex features such as lookbehind.

### Possible Drawbacks

Enabling PCRE2 potentially incurs performance penalty, as explained on [ripgrep's wiki](https://github.com/BurntSushi/ripgrep/blob/7b3fe6b3251a18d2b8d3efe7ee6a85c9e9e4e565/FAQ.md#why-does-ripgrep-get-slower-when-i-enable-pcre2-regexes), thus this PR does not enable it by default, and instead adds an option.

### Applicable Issues

Addresses https://github.com/atom/find-and-replace/issues/571

Accompanying [PR in atom core](https://github.com/atom/atom/pull/19615)

### DIFF
--- a/lib/project/results-model.js
+++ b/lib/project/results-model.js
@@ -217,6 +217,7 @@ module.exports = class ResultsModel {
 
     const startTime = Date.now()
     const useRipgrep = atom.config.get('find-and-replace.useRipgrep')
+    const enablePCRE2 = atom.config.get('find-and-replace.enablePCRE2')
 
     this.inProgressSearchPromise = atom.workspace.scan(
       this.regex,
@@ -225,6 +226,7 @@ module.exports = class ResultsModel {
         onPathsSearched,
         leadingContextLineCount,
         ripgrep: useRipgrep,
+        PCRE2: enablePCRE2,
         trailingContextLineCount
       },
       (result, error) => {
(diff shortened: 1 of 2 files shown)


dwelle:
Just reminding that this can now be merged, too.

rafeca:
Thanks for the reminder 😃

dwelle:
Mhm, seems it didn't work. I have both `ripgrep` and `PCRE2` option enabled, but the parser complains the regex is invalid. Is that coming from `ripgrep` or atom? How come specs passed?

image

Arcanemagus:
The [update of `find-and-replace`](https://github.com/atom/atom/commit/2b552482dd3dfd87ce9da3c94f2923a6651d9515) hasn't hit the beta channel yet, you can test it out on the [nightly releases](https://blog.atom.io/2019/06/21/introducing-atom-nightly-releases.html) now, or wait for it to hit beta/stable during the normal release cycle.

Issue #1097, opened 2019-09-04
Project Find Result not mantain scroll position after open search file


abcreations:
**Description**
Clicking search result in Project Find Result and open file, when go back to Project Find Result the scroll position is lost.

**Steps to Reproduce**
Perform a project-wide search for "foo"
Project Find Results opens
Scroll bottom and click a search result in Project Find Results
The file opened
Go back to Project Find Results

**Expected behavior:**
After go back to Project Find Result the scroll need to mantain position.

**Actual behavior:**
After go back to Project Find Result the scroll position are lost.

**Reproduces how often:**
After update to version 0.218.15

**Versions**
Atom : 1.40.1
Electron: 3.1.10
Chrome : 66.0.3359.181
Node : v10.2.0

Windows 10

calebmeyer:
I was able to reproduce this on my mac as well, so it looks like it's not OS specific. 

As a workaround, you can open up the settings for the find and replace package and set "Direction to open results pane" to right or down. I have mine set to right. I don't like losing the horizontal space, but it is nice to be able to click around between results.

JeffAsSilent:
Thanks for the workaround !
I have the same issue for some days (same versions as @abcreations)

adrian5:
Confirmed on Linux as well. Thanks for the workaround @calebmeyer ! I hope the issue gets some attention, and I suspect it's not a difficult fix.

PR #1105, opened 2019-11-08
Maintain previous scroll position

darangi:
### Description of the Change

Maintains scroll position after opening a file from the results page.

### Benefits

Always keep your last selected search result in focus whenever you return to the results page.

### Possible Drawbacks

None

### Applicable Issues

Fixes https://github.com/atom/find-and-replace/issues/1097


darangi:
> @darangi all looking good, just make sure the test fails without your fix and you are good to go on merging this PR. Good job 🤩
Thanks! The test fails without the fix 😃

darangi:
@lkashef  Thanks! The test fails without the fix 😃

MadLittleMods:
Can this be pushed into a new release? ✌😇 I'm stuck scrolling over and over

Aerijo:
@MadLittleMods I don't know where this particular PR is, but try out the beta or nightly channels if you want to use features sooner.

MadLittleMods:
@Aerijo The last release `v0.219.1` is from 2019-11-1. How do I switch to the beta/nightly of the `find-and-replace` package? I don't see any beta tags there. I'm just assuming the `find-and-replace` package is updated independent of Atom releases?

Or are you referring to the Atom as a whole nightlies? https://blog.atom.io/2019/06/21/introducing-atom-nightly-releases.html

Aerijo:
Ah, in that case it hasn't even come to nightly yet (Atom as a whole). What you can do now is clone the repo, run `apm install` and then `apm link` at it's root directory to use the repo in it's current state. I believe the install step will require build tools installed on your computer, for any native modules. The linked package will take priority over the builtin version (after a reload), and running `apm unlink find-and-replace` will remove this linked version and the builtin will be used again.

danialmuqris98:
Tq

mikkovedru:
Doesn't this commit also fixes https://github.com/atom/atom/issues/20259 ?

nobodo:
Not sure if related to this change, but a new quirk has appeared. When there are plenty of search results, collapse all, scroll down beyond the visible ones, then expand one and follow it, return back to search tab and then collapse that item -> position jumps. After this happens, any expand/collapse event jumps the position. The only cure is to close the search tab and do the search again (unfortunately the same thing happens again, but on the positive side this should be quite easy to reproduce)

Might not happen at the very first iteration, but I can reproduce this after few expand & follow & return. Happens also with nightly build (1.48.0-nightly2). Platform is macOS.

stwr667:
I get this too @nobodo. Really annoying when you're working through a big search list.

Issue #360, opened 2015-03-05
Option to have Find in Project open in a new tab each invocation


tkieft:
I often want to have the results of multiple "find in project" commands open at once. This is currently not possible.

bronson:
Me too!  This gets in my way a few times a week, really frustrating.

#### the problem
If I'm working with gigantic set of search results that took 20 seconds to compute, find-and-replace's current design prevents me from running any more searches because that would blow my hard-won results away.

If I could run another search, opening its results in a new Find Results tab, then I could keep the old results and still run new searches.

#### potential solution

It might not need to be an option though...  How about placing a "New Results Tab" button in the Project Find Results page?  Clicking this would open a new tab and steer all search results there.  The existing tab would be safe from modification until you close it.

Another solution could be Vim's trick of saving previous search results and allowing you to jump backward and forward at any time.  But this is overcomplex, slower, and seems less useful than just allowing you to open a new tab any time you have results that you want to save.

SebastianZaha:
+1 it is necessary when refactoring / reviewing larger changes especially in dynamic languages.

Gitbg:
Yes, I think the priority of the issue should be higher !

rubencaro:
Big +1 for this

rymohr:
Has my vote too.

daggeto:
:+1:

gigmaps:
+1

lodalo:
+1

bendoane:
+1...thousand. This would be game-changing for anyone working in large code bases. If I'm searching throughout the code and need to compare a few files found via different searches, and the files are in very different locations, the multiple tabs ability for the search would save SO much time and make this process easier.

cchatham:
👍

sksea:
+1

ChaosSteffen:
+1

chikamichi:
I believe it relates to [pending pane items](http://blog.atom.io/2016/03/17/atom-1-6-and-1-7-beta.html#pending-pane-items) that got introduced a few versions back.

When one clicks on a search result, the tab which is opened has its title in *italic*: it means the tab is "transient" or "pending", and it will get replaced if another search result is selected. Simply double-click on the tab's title to freeze it (its title will turn to regular, it's not pending anymore). The next time a search result is selected, a new, pending tab will be opened.

[...]

PR #1109, opened 2019-12-15
Add option to save search resulst from being overwrittenby next search

Indieberrie:
### Description of the Change

Every project search, overrides last search's results. This PR adds button to results tab, that when clicked, saves that tab, so it won't be overwritten by next search. Next search will open results in new tab (that can be overwritten, unless button in that one is clicked). After clicking button gets disabled

image
image

### Alternate Designs

Main focus is to have multiple results open at the same time. Other approach was to put this button, but it was easy to lose track of what tabs would stay open, and what overwritten.
image

### Benefits

Easier code navigation

### Applicable Issues

https://github.com/atom/find-and-replace/issues/360

### DIFF
--- a/lib/find.coffee
+++ b/lib/find.coffee
@@ -21,7 +21,7 @@ module.exports =
     atom.config.unset('find-and-replace.openProjectFindResultsInRightPane')
 
     atom.workspace.addOpener (filePath) ->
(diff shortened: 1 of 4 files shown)


darangi:
Thanks for the contribution 🙇  @Indieberrie

PR #897, opened 2017-05-17
Added prev button

jonathan-dejong:
### Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* All new code requires tests to ensure against regressions

### Description of the Change

Simply added the Find prev button back into the find and replace view.

### Alternate Designs

No other alternate designs. People want the button.

### Benefits

A new button called "Find prev" will be added before the Find button.

### Possible Drawbacks

A change in UI that might stump people used to having the "Find" button first. I added the "Find prev" button before because it's the logical location.

### Applicable Issues

None.

PR #1112, opened 2020-02-17
add find and replace previous buttons

### LINKED ISSUE #269: Add select-previous command
Seems like there ought to be a `find-and-replace:select-previous` to go along with the current `find-and-replace:select-next` command.
I am trying to mimic some behavior from Ace and I'm unable to without this.

### LINKED ISSUE #453: Option: Previous button visibility on ctrl-F
It would be very nice to have an option that lets the user decides if the previous button is visible or not in the ctrl-F window.

I have read previous issues - all closed - relating to this situation.

In all of them, I have the feeling this request has not been taken seriously. For example, duplicated issues demanding the previous button have been closed as duplicates. But never is there any mention that it will be done in the future. What bothers me most is that the main issue (https://github.com/atom/find-and-replace/issues/322) is not even related to any development request, so all fo those request are not even duplicates.
But that is out of the point, what matters the most is the option to add the previous button.

First of all, I would like to question the assumption you had when you explained that very few people use that button. How did you realized that? How many people were asked? It simply seems like a design decision since I hardly see how you could place the previous button in the interface.

In my opinion, the previous button is standard for an IDE and removing it would cause questioning and frustration to the user.

Also, If you plan on programming this, please leave this open for tracking, otherwise It feels like you're taking this suggestion to the trash.

Thank you

### LINKED ISSUE #467: Allow navigating to previous results
I often want to see a previous project find's results...  Maybe 'back' and 'forward' buttons could be added to the Project Find Results tab?  That way I can do a big search, do a quick search, and then hit the back button to go back to my big search results.  Later I'd hit forward to go back to the quick search, and so on.

It would make it much easier to chase down rabbit holes using Atom.

I probably want this because vim has :cold and :cnew commands ([description](http://thepugautomatic.com/2014/03/stacked-vim-searches-down-cold/)) -- very useful!

This might be related to https://github.com/atom/find-and-replace/issues/360.  Implementing one would reduce the need for the other but I hope both could be provided one day.

Tag me enhancement!

### LINKED ISSUE #800: Add select previous
Given I have a word selected like so:

One word, two word, red word, blue `word`

I would like a keyboard shortcut that adds the immediately prior instance of that word to the selection, so the state after hitting that shortcut would be:

One word, two word, red `word`, blue `word`

This would be the opposite of the current `find-and-replace:select-next` feature.

Issue #324 seems to suggest that this feature was available at one point but since there is not a reference to it in the current [menu](https://github.com/atom/find-and-replace/blob/e4fca61c9ec8b339c76686c4efa76aae5e9840a1/menus/find-and-replace.cson) or [keymap](https://github.com/atom/find-and-replace/blob/e4fca61c9ec8b339c76686c4efa76aae5e9840a1/keymaps/find-and-replace.cson) cson files, it seems to have been removed at some point? Or was that referring to the distinct feature `find-and-replace:find-previous` that is currently bound by default to `cmd-shift-g`?

Using that feature in from my initial example state results in the following:

 One word, two word, red `word`, blue word

which is not what I want.

UziTech:
### Description of the Change

Add `Find Previous` and `Replace Previous` buttons to view

image

### Alternate Designs

We could actually write words on the buttons but I feel like it is intuitive with the arrows. There is also tooltips on each button that says what it does so someone could hover over the button to see what it does.

image

### Benefits

Easily find/replace previous from view. I know <kbd>shift</kbd> + <kbd>Find</kbd> finds the previous element but it is not documented anywhere and there is no equivilant <kbd>shift</kbd> + <kbd>Replace</kbd>

### Possible Drawbacks

none

### Applicable Issues

closes #897

fixes #269
fixes #453
fixes #467
fixes #800

### DIFF
--- a/lib/find-view.js
+++ b/lib/find-view.js
@@ -79,8 +79,12 @@ class FindView {
           ),
 
           $.div({className: 'input-block-item'},
-            $.div({className: 'btn-group btn-group-find'},
-              $.button({ref: 'nextButton', className: 'btn btn-next'}, 'Find')
+            $.div({className: 'btn-group btn-group-find-prev'},
+              $.button({ref: 'prevButton', className: 'btn btn-prev'}, '<')
+            ),
+
+            $.div({className: 'btn-group btn-group-find-next'},
+              $.button({ref: 'nextButton', className: 'btn btn-next'}, '>')
             ),
 
             $.div({className: 'btn-group btn-group-find-all'},
@@ -100,8 +104,12 @@ class FindView {
           ),
 
           $.div({className: 'input-block-item'},
-            $.div({className: 'btn-group btn-group-replace'},
-              $.button({ref: 'replaceNextButton', className: 'btn btn-next'}, 'Replace')
+            $.div({className: 'btn-group btn-group-replace-prev'},
+              $.button({ref: 'replacePrevButton', className: 'btn btn-prev'}, '<')
+            ),
+
+            $.div({className: 'btn-group btn-group-replace-next'},
+              $.button({ref: 'replaceNextButton', className: 'btn btn-next'}, '>')
             ),
 
             $.div({className: 'btn-group btn-group-replace-all'},
@@ -223,7 +231,8 @@ class FindView {
 
   handleEvents() {
     this.findEditor.onDidStopChanging(() => this.liveSearch());
-    this.refs.nextButton.addEventListener('click', e => e.shiftKey ? this.findPrevious({focusEditorAfter: true}) : this.findNext({focusEditorAfter: true}));
+    this.refs.prevButton.addEventListener('click', () => this.findPrevious({focusEditorAfter: true}));
+    this.refs.nextButton.addEventListener('click', () => this.findNext({focusEditorAfter: true}));
     this.refs.findAllButton.addEventListener('click', this.findAll.bind(this));
     this.subscriptions.add(atom.commands.add('atom-workspace', {
       'find-and-replace:find-next': () => this.findNext({focusEditorAfter: true}),
@@ -235,6 +244,7 @@ class FindView {
       'find-and-replace:use-selection-as-replace-pattern': this.setSelectionAsReplacePattern.bind(this)
     }));
 
+    this.refs.replacePrevButton.addEventListener('click', this.replacePrevious.bind(this));
     this.refs.replaceNextButton.addEventListener('click', this.replaceNext.bind(this));
     this.refs.replaceAllButton.addEventListener('click', this.replaceAll.bind(this));
     this.subscriptions.add(atom.commands.add('atom-workspace', {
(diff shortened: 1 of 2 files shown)


UziTech:
I could also add a setting to enable the buttons.

find-and-replace

UziTech:
The buttons could be arrows or any other icon from [octicons](https://octicons.github.com/)

image

diveshkr-code:
Hey I feel 'replace next' and 'replace previous' seems not very useful and dangerous options, for example, suppose I mistakenly clicked 'replace previous' arrow and didn't realize that I have done so how will I get it back. 
Hence I suggest that only 'find next' and 'find previous' should be added.

UziTech:
@diveshkr-code  <kbd>Ctrl</kbd> + <kbd>z</kbd> will undo the replace

UziTech:
@atom @darangi @nathansobo any feedback on this?

diveshkr-code:
> @diveshkr-code kr-code Ctrl + z will undo the replace

Surely@UziTech but what if the user clicks 'replace previous' without realizing he has done it and he doesn't see it change as the previous occurrence was not visible on the screen. 
Since the code is changed with replace option I think it's better if replace is done only on the current selection. 
Just my opinion. Cheers.

UziTech:
@diveshkr-code There are many things the user can do to accidentally change something, but I don't think we should remove the functionality.

For example, my cat walked on my keyboard the other day and entered a bunch of random text as well as pushing the page down button so I couldn't see what was changed. That doesn't mean we should remove the ability to move down a page with the page down button.

The "Replace Previous" and "Replace Next" buttons are extremely useful when renaming a variables or replacing other text only in certain places. Some of the found text you might want to not replace so you can click "Find Next" and some you will want to replace so you can click "Replace Next".

UziTech:
@diveshkr-code This PR isn't adding any new functionality. The current "Replace" button does a replace and finds the next occurrence (the same thing the replace next button will do in this PR), and the ability the replace previous is in the command palette. This PR just adds buttons for Find/Replace Previous

diveshkr-code:
> @diveshkr-code This PR isn't adding any new functionality. The current "Replace" button does a replace and finds the next occurrence (the same thing the replace next button will do in this PR), and the ability the replace previous is in the command palette. This PR just adds buttons for Find/Replace Previous

@UziTech I think you have me pretty convinced cheers.

UziTech:
@atom @darangi @rafeca @nathansobo Who is reviewing this?

darangi:
> @atom @darangi @rafeca @nathansobo Who is reviewing this?

@UziTech, this has been added to our radar, the team will get back to you soonest. Thanks for the contribution 🙇‍♂️

sadick254:
Hey @UziTech,
This PR adds a UI to a functionality that is already present within atom. When user opens `find-and-replace` settings, the user can see the shortcuts to find/ replace next and previous matches. The benefits we get from having a UI behind that functionality is very minimal. Thank you for the effort to improve find and replace, unfortunately we will have to close this one out.

UziTech:
>  The benefits we get from having a UI behind that functionality is very minimal.

That is very subjective which is why we could place this behind a setting checkbox.

If you don't think Atom will allow this type of UI you should also close the issues this is PR would have fixed so people don't go there thinking this is something that will be allowed.

#897, #269, #453, #467, #800

Blosberg:
I'm struggling to understand your rationale here @sadick254.

The fact that there is, by default, no "Find previous" button is just baffling, and @UziTech has done the work to implement it, so there's no workload issue here. I can't understand the conscious decision to omit such an absolutely fundamental piece of functionality when it costs nothing. 

The tragedy here is that I love everything else about Atom, but this is a deal-breaker. It's like you've built a Ferrari, and then removed the "reverse" gear, telling people they need to memorize some strange code just to back up. You say "the benefits are minimal" (to which I disagree, I think the benefits are minimal, absolutely critical, functionality), but what exactly is the cost? The work is done, there's plenty of space, there are clearly a lot of people who want this functionality. Is there any actual reason to not implement this?

Issue #398, opened 2015-05-27
Multi-line find and replace


solomonkinard:
The ability to find and replace multiple lines would be epic!

#397 is different from #396 because it is requesting multiple rows in both the find and replace fields.
#396 only accounts for the insertion of a newline character in the replace field.

benogle:
Is this for buffer find or project find?

solomonkinard:
It could be useful in both, so feel free to choose the easiest one to implement (e.g. buffer).

NathanRSmith:
I would like to see this as well. Thanks.

ChrisDavison:
+1

dschwen:
This already works in buffer find (with regular expressions containing ```\n```), however it does **not** seem to work in project find.

yoyurec:
+1 for `\n`

dlewand691:
+++++++1111111 
Have to jump back to Sublime to do multiline find/replace across projects. LAME!

varemenos:
:+1:

ashtonwar:
+1

TW80000:
Just a heads up for anyone that might run into the same problems I did:

1. On Windows newlines are denoted by `\r\n`, not just `\n` like on unix-like systems. You'll need this for multi-line regex find expressions.
2. If you use the `$` character in a regex followed by your `\n` or `\r\n`, you'll be able to highlight text that matches the regex, but the replace will fail (in my experience). Taking out the `$` will allow you to do the replace.

mik01aj:
+1.

Related: https://discuss.atom.io/t/multi-line-find-and-replace/10222

johnjelinek:
:+1:

reduardo7:
+1

tlrobinson:
:+1:

borisschapira:
:+1:

leberblock:
+1

paullryan:
:+1:

quaritexa:
:thumbsup:

colin-kiegel:
+1

Ehres:
+1

pabloab:
+1

fmaj7:
+1

koltejas:
+1

hanhpho:
+1

vincentorback:
:+1:

hellpf:
+1

entr:
+1

pashast:
+1

co2-git:
[...]

Issue #396, opened 2015-05-27
Print newline character when Ctrl + Enter and/or Alt + Enter is pressed


ssorallen:
It's currently not possible to type a newline character in any of the mini TextEditors in find/replace, but pasting newline characters into any of them and then running the search works as expected. Pressing Alt / Ctrl + Enter currently does nothing.

If Ctrl or Alt is pressed with Enter, it'd be nice to print a newline character in the focused TextEditor.

I wasn't sure if this request belonged in find-and-replace, core, or atom-space-pen-views.

devios1337:
+1

Tab character as well please!

mjrk:
I'm very in favour of this feature, +1.

However, I would vote for Shift+Enter (and maybe redundantly/optionally Ctrl+Enter if a lot people vote for it) as I know it like this e.g. Empathy and other chat programmes. Alt+Enter would conflict with Windows console fullscreen (although I don't use windows anymore), Ctrl+Enter seems more like "committing/executing" than "not committing but inserting", e.g. Thunderbird Ctrl+Enter sends the email.

vjeux:
Here's Sublime behavior for this:

- In a normal file search, shift-enter goes to the previous occurrence, ctrl-enter and cmd-enter adds a \n
- In a find files search, shift-enter and ctrl-enter adds a \n, but cmd-enter runs the actual search.

PR #1118, opened 2020-03-27
Insert newline

thirstyice:
### Description of the Change

Add support for inserting newlines into find and replace fields using a keyboard shortcut (Cmd/Ctrl-Enter by default)

### Alternate Designs

Used Cmd/Ctrl-Enter for the default keyboard shortcut because it's functions tended to overlap with other shortcuts, making it redundant.

### Benefits

Ability to find + replace strings containing newlines without needing to copy/paste the newline into the text field

### Possible Drawbacks

Remapping of default keyboard shortcuts may affect existing workflows.

The code literally just allows for easier insertion of newlines into the text fields. How the program handles newlines is not affected, so if the program is unable to handle newlines in certain cases, this won't change that.

### Applicable Issues
Direct solution for #396 
Also affects #398

### DIFF
--- a/keymaps/find-and-replace.cson
+++ b/keymaps/find-and-replace.cson
@@ -34,7 +34,7 @@
 
 '.platform-darwin .find-and-replace':
   'shift-enter': 'find-and-replace:show-previous'
-  'cmd-enter': 'find-and-replace:confirm'
+  'cmd-enter': 'find-and-replace:insert-newline'
   'alt-enter': 'find-and-replace:find-all'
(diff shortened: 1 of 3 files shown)


sadick254:
Hey @thirstyice . Congratulations on your first contribution. Unfortunately find-and-replace doesn't support multiline searches because it works on a line by line basis. Inserting a new line on a search term would result in empty result. In order for us to accept this PR we would first need to add support for multiline searches. We currently don't have plans to add that feature.

PR #1137, opened 2020-09-30
Add devtools catch

Aerijo:
### Requirements

### Description of the Change

Reduces flaky test failures on the main Atom repo CI. The summary is that an uncaught exception is triggering dev tools, which breaks tests. See https://github.com/atom/atom/pull/21335 for more.

### Alternate Designs

Fixing the root of the problem. But that's a lot more difficult, seeing as the cause hasn't been completely determined yet. Investigations are welcome! 

### Benefits

Reduced false CI failures on main repo PRs

### Possible Drawbacks

This is only masking the deeper issue, so isn't a fix and should not be treated as such.

### Applicable Issues

None. It does not completely resolve any issues.

### DIFF
--- a/spec/setup-spec.js
+++ b/spec/setup-spec.js
@@ -2,3 +2,18 @@ require('etch').setScheduler({
   updateDocument(callback) { callback(); },
   getNextUpdatePromise() { return Promise.resolve(); }
 });
+
+// The CI on Atom has been failing with this package. Experiments
+// indicate the dev tools were being opened, which caused test
+// failures. Dev tools are meant to be triggered on an uncaught
+// exception.
+//
+// These failures are flaky though, so an exact cause
+// has not yet been found. For now the following is added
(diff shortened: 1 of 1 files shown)


spec/setup-spec.js
```diff
@@ -2,3 +2,18 @@ require('etch').setScheduler({
...
+// indicate the dev tools were being opened, which caused test
+// failures. Dev tools are meant to be triggered on an uncaught
+// exception.
+//
+// These failures are flaky though, so an exact cause
+// has not yet been found. For now the following is added
+// to reduce the number of flaky failures, which have been
+// causing false failures on unrelated Atom PRs.
+//
+// See more in https://github.com/atom/atom/pull/21335
```
sadick254, line 16: Thanks for adding this context @Aerijo

sadick254 (approved):

PR #1153, opened 2021-05-09
Add font resizing in project search (#21125)

dawidowoc:
### Requirements

### Description of the Change

Add possibility to change font size in project find results view. The font size is connected to _editor.fontSize_ property. It can be changed by using _window:increase-font-size_, _window:decrease-font-size_, _window:reset-font-size_ commands, similarly as you can do for editor view.

Video presenting the feature: https://youtu.be/fqGRfRE3b6M

### Alternate Designs

I don't see any alternate designs.

### Benefits

Possibility to change font size in project find results view.

### Possible Drawbacks

I don't see any.

### Applicable Issues

https://github.com/atom/atom/issues/21125

### DIFF
--- a/lib/project/list-view.js
+++ b/lib/project/list-view.js
@@ -14,6 +14,22 @@ module.exports = class ListView {
     const resizeObserver = new ResizeObserver(() => etch.update(this));
     resizeObserver.observe(this.element);
     this.element.addEventListener('scroll', () => etch.update(this));
+
+    this.registerFontResizing();
+  }
+
+  registerFontResizing() {
+    atom.commands.add('atom-workspace', {
+      'window:increase-font-size': () => {
+        etch.update(this)
+      },
+      'window:decrease-font-size': () => {
+        etch.update(this)
+      },
+      'window:reset-font-size': () => {
+        etch.update(this)
+      }
+    })
   }
 
   update({items, heightForItem, itemComponent, className} = {}) {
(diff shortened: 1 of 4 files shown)


lib/project/results-view.js
```diff
@@ -96,18 +96,28 @@ class ResultsView {
...
-      return this.matchRowHeight
+      baseHeight = this.matchRowHeight
     }
+
+    return this.getHeightAdjustedToFontSize(baseHeight)
+  }
+
+  getHeightAdjustedToFontSize(baseHeight) {
+    const lineHeight = 2;
+    const baseFontSize = 12;
```
dawidowoc, line 116: 12 was the default font size before my changes. I'm treating it as base size for height calculation because some of the existing stylings seems to depend on this value.

lib/project/results-view.js
```diff
@@ -96,18 +96,28 @@ class ResultsView {
...
     } else if (resultRow instanceof MatchRow) {
-      return this.matchRowHeight
+      baseHeight = this.matchRowHeight
     }
+
+    return this.getHeightAdjustedToFontSize(baseHeight)
+  }
+
+  getHeightAdjustedToFontSize(baseHeight) {
+    const lineHeight = 2;
```
dawidowoc, line 115: Value of line-height for items in project find results is equal to 2em. That's why we need to keep this multiplier when increasing/decreasing font size.

lib/project/list-view.js
```diff
@@ -54,18 +70,20 @@ module.exports = class ListView {
         children.push(
           $.div(
             {
+              className: "item-container",
```
dawidowoc, line 73: Needed for selection during test only.

styles/find-and-replace.less
```diff
@@ -374,3 +374,7 @@ atom-workspace.find-visible {
     }
   }
 }
+
+.list-tree.has-collapsable-children .list-nested-item > .list-item::before {
+  vertical-align: middle;
```
dawidowoc, line 379: It's related to that arrow icon showing that you can expand a nested item. We want them to keep it in the middle (vertically) at all times when you change the font size.

